PR #12, opened 2018-10-04
Logo Design

Quantumf0x:
I made a logo design from your sketch


Quantumf0x:
Oke i was thinking you were oke that i tweaked it.

NullVoxPopuli:
I mean, what you have is _much_ cleaner than what I did. So that's def a plus. :+1:
I'm looking for a different design entirely though.

PR #84, opened 2019-05-28
Remove duplicated contact entries

brunoocasali:
And add a little context separation of what would be "set" and "unload".


NullVoxPopuli:
thanks a bunch! this is a good change :)

PR #109, opened 2019-09-04
Build Work / Path to Embroider

### LINKED ISSUE #47: For a conversation, the new messages since floater does not appropriately disappear when messages are read
There needs to be tests for this feature, anyway

### LINKED ISSUE #49: When looking at a conversation, messages received remain unread until clicked.
If the window has focus, they should be marked as read.

### LINKED ISSUE #76: When using ctrl+k + enter to select a user/channel to jump to, focus is no where to be found. The focused element should be the chat entry input field.
I don't even know where the focus is, so I don't know how long it'd take to tab to the chat input.
I wonder if it makes sense to have a hotkey to focus the active chat's input. 
Could be useful if you're focused on something else (such as an image, code snippet, etc), and want to jump to the input to send a message.

NullVoxPopuli:
Resolves #76 
Resolves #49
Resolves #47?


NullVoxPopuli:
Travis Passed, merging soon. Will have to do a manual deploy, since I don't have continuous deployments set up yet.

Issue #29, opened 2018-11-23
Use `node-emojis` for emoji name replacement


NullVoxPopuli:
https://github.com/tallesl/node-emojis
should this be live-swapped as you type? yes.
bundle size: https://bundlephobia.com/result?p=emojis@1.0.10 (small)

Add an option to disable

tonowoe:
Yes! 😂

NullVoxPopuli:
Resolved in #181

PR #181, opened 2019-10-02
Emoji support

amgassert:
Bare-bones prototype for emoji support

Addresses #121 and #29


packages/frontend/app/pods/components/pod/chat/chat-entry/emoji-parser.ts
```diff
@@ -0,0 +1,23 @@
+const emojiDict = new Map([
+    [":partyparrot:", "'(\\(0>0)/)'"] // TODO: use gif
+]);
+
+export default class EmojiParser {
```
NullVoxPopuli: because this class doesn't maintain any state, it could probably be converted into a pure function

packages/frontend/app/pods/components/pod/chat/chat-entry/emoji-parser.ts
```diff
@@ -0,0 +1,23 @@
...
+
+    parseEmoji(message: string): string {
+        let emojiRegex = /:.*:/gi;
```
NullVoxPopuli: .* includes `:`

what happens when you have `this :+1: text :-1:`?

packages/frontend/app/pods/components/pod/chat/chat-entry/emoji-parser.ts
```diff
@@ -0,0 +1,23 @@
...
+        let matches = message.match(emojiRegex);
+        let emojifiedMessage : string = message;
+    
+        if (matches !== null) {
```
NullVoxPopuli: this whole block could maybe be simplified a bit by using the function replacer variant of the `replace` method?
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replace#Using_an_inline_function_that_modifies_the_matched_characters

agassert:
To-do:
- [x] Convert emoji parser class to pure function
- [ ] Fix regex in order to support multiple matches
- [ ] Simplify emoji replacement with function replacer variant of `replace` method

NullVoxPopuli:
I see you have a resolve merge conflicts commit
image


I don't think there should be any conflict resolution commits? how'd did you get there?

here is what I typically do:

```
git checkout master
git fetch origin
get reset origin/master --hard
git checkout emoji-support
git rebase master
# if conflicts, resolve
git add <resolved file>
git rebase --continue
git push origin emoji-support --force-with-lease
```

jsconfig.json
```diff
@@ -0,0 +1 @@
+{"compilerOptions":{"target":"es6","experimentalDecorators":true},"exclude":["node_modules","bower_components","tmp","vendor",".git","dist"]}
```
NullVoxPopuli: This file shouldn't exist?

.gitignore
```diff
@@ -5,6 +5,7 @@ _build
 deps
 *.log
 .idea/
+.DS_Store
```
NullVoxPopuli: :tada: :balloon:

packages/frontend/app/pods/components/pod/chat/chat-entry/component.ts
```diff
@@ -59,6 +60,7 @@ export default class ChatEntry extends Component<IArgs> {
     if (!this.text) return;
 
     this.isDisabled = true;
+    this.text = unicode(this.text);
```
NullVoxPopuli: to make the unicode live update as you type, you could invoke the unicode function on line 84

packages/frontend/app/pods/components/pod/chat/chat-entry/component.ts
```diff
@@ -79,7 +80,11 @@ export default class ChatEntry extends Component<IArgs> {
   @action onInput(event: KeyboardEvent) {
     const value = (event.target as any).value;
 
-    this.text = value;
+    if (value.charAt(value.length - 1) === ':') {
```
NullVoxPopuli: what happens when you go back and add an emoji somewhere in the middle of your text?

agassert: Fixed, thanks!

packages/frontend/app/pods/components/pod/chat/chat-entry/component.ts
```diff
@@ -77,9 +78,14 @@ export default class ChatEntry extends Component<IArgs> {
   }
 
   @action onInput(event: KeyboardEvent) {
+    const EMOJI_REGEX = /:[^:]+:/g;
     const value = (event.target as any).value;
 
-    this.text = value;
+    if (value.match(EMOJI_REGEX)) {
```
NullVoxPopuli: based on https://github.com/NullVoxPopuli/emberclear/pull/242

I think it's fast enough where we actually don't need to wrap this in a condition.

Also, I learned this today:
image
So.. that's weird -- the regex caches the value and becomes a toggle

NullVoxPopuli: ah, but you're using match

packages/frontend/app/pods/components/pod/chat/chat-entry/component.ts
```diff
@@ -77,9 +78,14 @@ export default class ChatEntry extends Component<IArgs> {
   }
 
   @action onInput(event: KeyboardEvent) {
+    const EMOJI_REGEX = /:[^:]+:/g;
```
NullVoxPopuli: Should this regex be pulled out of the class, since it seems to be static?

packages/frontend/tests/integration/components/pod/chat/chat-entry/emoji-test.ts
```diff
@@ -0,0 +1,70 @@
...
+import { hbs } from 'ember-cli-htmlbars';
+
+module('Integration | Component | chat-entry', function(hooks) {
+  setupRenderingTest(hooks);
+
+  module('emoji code replacement', function() {
+    module('there are no emoji codes to replace', function() {
+      test('result does not contain emoji', async function(assert) {
+        assert.expect(2);
+        await render(hbs`<Pod::Chat::ChatEntry />`);
```
NullVoxPopuli: Since both of these tests share the same first 4 lines, how do you feel about the render, and empty check/assert getting pulled up into a before each?

Reduce duplication, and results in what you're actually testing being more focussed

packages/frontend/app/pods/components/pod/chat/chat-entry/component.ts
NullVoxPopuli: thoughts on this?

packages/frontend/tests/integration/components/pod/chat/chat-entry/emoji-test.ts
```diff
@@ -0,0 +1,88 @@
...
+import { setupRenderingTest } from 'ember-qunit';
+import { hbs } from 'ember-cli-htmlbars';
+import { getStore } from 'emberclear/tests/helpers';
+
+module('Integration | Component | chat-entry', function(hooks) {
+  setupRenderingTest(hooks);
+
+  module('emoji code replacement', function() {
+    module('there are no emoji codes to replace', function() {
+      test('result does not contain emoji', async function(assert) {
```
NullVoxPopuli: because these two tests share the same setup, should that be extracted to a `beforeEach`?

packages/frontend/tests/integration/components/pod/chat/chat-entry/emoji-test.ts
```diff
@@ -0,0 +1,88 @@
+import { module, test } from 'qunit';
+import { render, fillIn } from '@ember/test-helpers';
+import { setupRenderingTest } from 'ember-qunit';
+import { hbs } from 'ember-cli-htmlbars';
+import { getStore } from 'emberclear/tests/helpers';
+
+module('Integration | Component | chat-entry', function(hooks) {
```
NullVoxPopuli: there is one more test missing from this:
 - where we enter text, emojis get replaced, and then we enter more text, and more emojis get replaced.

packages/frontend/app/pods/components/pod/chat/chat-entry/component.ts
amgassert: Seems reasonable.

- [ ] Pull out emoji regex

packages/frontend/tests/integration/components/pod/chat/chat-entry/emoji-test.ts
amgassert: Done

packages/frontend/app/pods/components/pod/chat/chat-entry/component.ts
amgassert: Seems reasonable

packages/frontend/tests/integration/components/pod/chat/chat-entry/emoji-test.ts
amgassert: Done

PR #187, opened 2019-10-05
Build using embroider

NullVoxPopuli:


packages/frontend/package.json
```diff
@@ -16,8 +16,9 @@
     "build": "yarn ember build",
     "build:production": "EXPERIMENTAL_RENDER_MODE_SERIALIZE=true yarn build --environment production && sed -i -e 's/{{ROOT_URL}}/\\//g' dist/index.html",
     "build:production:analyze": "CONCAT_STATS=true yarn build:production",
-    "start": "yarn ember serve -p 4201",
-    "start:dev": "cross-env SW_DISABLED=true yarn start",
+    "start:sw": "yarn ember serve -p 4201",
```
villander: I think that you need install embroider via package.json

NullVoxPopuli: It's already in there ;)

NullVoxPopuli: https://github.com/NullVoxPopuli/emberclear/pull/187/files#diff-e2f6b05cb1df8e9881d391bf6c0c0beeR57

Issue #46, opened 2018-11-27
Update the page title to show the current total number of unread messages


NullVoxPopuli:
Just so you can see how many unread messages you have if you have a lot of tabs open.

`34 - emberclear - NullVoxPopuli` could be an example title.

CelticMajora:
A few related links from quick research:

- https://www.balinterdi.com/blog/setting-the-document-title-in-ember-apps/
- https://github.com/kimroen/ember-cli-document-title

CelticMajora:
Addressed by #226

PR #226, opened 2019-10-07
Include unread message count in tab title

CelticMajora:
Addresses Issue #46


CelticMajora:
Currently not working. Want to do some more research on ember conventions to make sure there isn't a more standard way that would actually work. I think the top lay index.html title is just overriding the top-nav title currently.

NullVoxPopuli:
> Want to do some more research on ember conventions to make sure there isn't a more standard way that would actually work. 

For now, it'll be tricky to find anything, because ember is going through a shift in the mental model.

As we work on this project, we are pioneering / setting an example. :)

packages/frontend/app/pods/components/pod/application/top-nav/template.hbs
```diff
@@ -1,3 +1,4 @@
+<head><title>emberclear {{this.unreadMessageCount}} {{this.currentUser.name}}</title></head>
```
NullVoxPopuli: This looks like a good way to format the title, however:

The head tag may only live within the HTML tag's content. All app code lives in the body tag.

Since we can set the title in JavaScript via document.title, we should just do that. Don't need have this kind of thing in the template.

Just need a render modifier to call a function that set's the title

NullVoxPopuli:
A couple notes from discord:

> I think the best approach is via a modifier that extends ember-class-based-modifier that attaches itself to https://github.com/NullVoxPopuli/emberclear/blob/master/packages/frontend/app/pods/components/pod/application/top-nav/template.hbs#L12
> {{update-document-title this.doTheUpdate this.messages}}

CelticMajora:
Is this the documentation for the modifiers you're talking about?
https://www.npmjs.com/package/ember-class-based-modifier

NullVoxPopuli:
@CelticMajora yup! you that's exactly the docs/readmes/blogposts I'm talkin' about

NullVoxPopuli:
based off the commit history, it looks like you'll want to rebase on upstream/master.

to do that:
```
git remote add upstream git@github.com:NullVoxPopuli/emberclear.git
git fetch upstream
git checkout master
git checkout feature-unread-message-count-title
git rebase master
```

CelticMajora:
Awesome. I have the remote set up already, so I'll update my master branch and this branch

packages/frontend/app/modifiers/update-document-title.ts
```diff
@@ -0,0 +1,26 @@
...
+  get currentUserName(): string {
+    return <string>this.args.named.currentUserName;
+  }
+
+  willDestroy() {
+    document.title = this.originalDocumentTitle;
+  }
+
+  didReceiveArguments() {
+    document.title = 'emberclear ' + this.unreadMessageCount + ' ' + this.currentUserName;
```
NullVoxPopuli: It's a smidge more concise, in JS/TS to do `emberclear ${this.unreadMessageCount} ${this.currentUserName}`

NullVoxPopuli: do we want the unread count first if the tab is pinned?

packages/frontend/app/pods/components/pod/application/top-nav/template.hbs
```diff
@@ -2,14 +2,14 @@
 
   <div class='left-nav'>
     {{#if this.currentUser.isLoggedIn}}
-      <div>
+      <div {{update-document-title unreadMessageCount=this.unreadMessageCount currentUserName=this.currentUser.name}}>
```
NullVoxPopuli: I like where you're head's at with the readability here! it's very clear what argument is being used for what.
however, the currentUser doesn't need to be passed, because we can access it directly from the modifier, like so:

```ts
import { inject as service } from '@ember/service';

// ...

export default class ... extends ...  {
  @service currentUser;

   // ...

   document.title = `${unreadCount} ${this.currentUser.name} .... `;
}
```

and given that, we would then only need one argument passed to update-document-title, and since there is only one, it could be a positional param without any lost semantics.

thoughts?

packages/frontend/app/modifiers/update-document-title.ts
```diff
@@ -0,0 +1,26 @@
...
+
+export default class UpdateDocumentTitle extends Modifier {
+  originalDocumentTitle: string;
+
+  constructor(owner, args) {
+    super(owner, args);
+    this.originalDocumentTitle = document.title;
+  }
+
+  get unreadMessageCount(): number {
```
NullVoxPopuli: since these are only used in the one method, do they need to be defined as getters?

packages/frontend/app/pods/components/pod/application/top-nav/template.hbs
CelticMajora: That is good for this current use. If there are future use cases that need to change the document title, it will need to be modified.

packages/frontend/app/modifiers/update-document-title.ts
CelticMajora: All the examples in the docs have those as getters. I think it may have something to do with didReceiveArguments being updated whenever an argument's value is updated. I could be wrong though.

CelticMajora:
The current code still isn't changing the document title. I'll try to get to some debugging tomorrow.

packages/frontend/app/pods/components/pod/application/top-nav/template.hbs
NullVoxPopuli: What other uses can you think of?

Over designing up front is common, and often leads to incorrect abstractions ;)

packages/frontend/app/modifiers/update-document-title.ts
NullVoxPopuli: There is no functional difference. It's just for readability. Since the docs include it as an example, I'm cool with it staying :)

NullVoxPopuli:
Looks like it is, but maybe backticks aren't being used for string interpolation?

NullVoxPopuli:
Screenshot_20191016-061539

packages/frontend/app/modifiers/update-document-title.ts
```diff
@@ -0,0 +1,25 @@
...
+    return this.args.named.unreadMessageCount as number;
+  }
+
+  willDestroy() {
+    document.title = this.originalDocumentTitle;
+  }
+
+  didReceiveArguments() {
+    document.title = '${this.unreadMessageCount} - emberclear - ${this.currentUser.name}';
```
NullVoxPopuli: Looks like an accidental usage of ' instead of `

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals

packages/frontend/app/pods/components/pod/application/top-nav/template.hbs
```diff
@@ -2,14 +2,14 @@
 
   <div class='left-nav'>
     {{#if this.currentUser.isLoggedIn}}
-      <div>
+      <div {{update-document-title unreadMessageCount=this.unreadMessageCount}}>
```
NullVoxPopuli: Since there is only one arg passed in, do we need to name it?

As a consumer of this API, I don't know if I care that the first arg is called unreadMessageCount. I know that whatever I pass will be the first token in the title.

So, maybe to your point about reusability, maybe the arguments passed should be only positional args, and in the modifier you could do something like 
```ts
let tokens = this.args.positional.join(' | ');
document.title =`${tokens} | emberclear`;
```

That would also allow you to easily pass in the name, as it would just be another token

NullVoxPopuli:
Overall, this is looking good. Biggest thing left are tests.

For this task, you'll probably want to add to the existing chat/privately-with acceptance tests

packages/frontend/app/pods/components/pod/application/top-nav/template.hbs
CelticMajora: I like that idea

packages/frontend/tests/acceptance/chat/privately-with/-acceptance-test.ts
```diff
@@ -22,6 +22,14 @@ module('Acceptance | Chat | Privately With', function(hooks) {
   setupApplicationTest(hooks);
   clearLocalStorage(hooks);
 
+  module('is not logged in', function(hooks) {
+    test('document.title is unchanged', async function(assert){
+      var titleBefore = document.title;
```
NullVoxPopuli: this test doesn't visit any pages / do actions, so `await settled()` maybe doesn't do anything?

What is the expected pre-login title?

CelticMajora: I believe it is currently just "emberclear" but I didn't want to code that in. Would the site for not logged in still be /chat/privately-with or should it just be the root emberclear site?

NullVoxPopuli: emberclear is static in every language, so using it is fine. but also, you can test strings via i18n, like:

`this.owner.lookup('service:i18n').t('appname')` -- this is defined in en.yml.
 - this is the preferred (but not yet widely adopted way of testing string values)

 > Would the site for not logged in still be /chat/privately-with or should it just be the root emberclear site?

`/chat/privately-with` is not visitable when logged out, a redirect will occur to `https://emberclear.io/setup/new` (which there are tests for, I think)

CelticMajora: > `/chat/privately-with` is not visitable when logged out, a redirect will occur to `https://emberclear.io/setup/new` (which there are tests for, I think)

Will that redirect be done with the `await settled()`?

NullVoxPopuli: maybe?

if you were to `visit` without `await`ing, then that's when you'd want to `await settled()`

packages/frontend/tests/acceptance/chat/privately-with/-acceptance-test.ts
```diff
@@ -22,9 +22,34 @@ module('Acceptance | Chat | Privately With', function(hooks) {
   setupApplicationTest(hooks);
   clearLocalStorage(hooks);
 
+  module('is not logged in', function(hooks) {
+    hooks.beforeEach(async function() {
+      QUnit.config.altertitle = false;
```
NullVoxPopuli: I was looking at the documentation for this, and it looks like it would need to go in test-setup.js

packages/frontend/tests/acceptance/chat/privately-with/-acceptance-test.ts
```diff
@@ -21,9 +21,32 @@ module('Acceptance | Chat | Privately With', function(hooks) {
   setupApplicationTest(hooks);
   clearLocalStorage(hooks);
 
+  module('is not logged in', function(hooks) {
+    hooks.beforeEach(async function() {
+      await visit('/chat/privately-with');
+    });
+
+    test('document.title is unchanged', async function(assert) {
```
NullVoxPopuli: async is not needed here 
(I need to figure out what linter this is that will warn of that (no async when there is no await))

packages/frontend/app/pods/components/pod/application/top-nav/template.hbs
```diff
@@ -2,14 +2,14 @@
...
-      <div>
+      <div {{update-document-title this.unreadMessageCount}}>
         <HamburgerButton
           @onClick={{this.sidebar.toggle}}
           @isActive={{this.sidebar.isShown}}
         />
 
         {{#if this.hasUnread}}
-          <span
+          <span
```
NullVoxPopuli: looks like your editor may need a trim trailing whitespace option enabled :)

NullVoxPopuli:
This looks good. Having played with this, I think I'm going to merge it, and then open another issue for little tweaks and nice things we can add that will supplement the work you did here :D <3 thanks!

PR #244, opened 2019-10-09
Add Linting to Smoke Tests

NullVoxPopuli:
supersedes: https://github.com/kellyselden/emberclear/pull/1

### DIFF
--- a/packages/smoke-tests/package.json
+++ b/packages/smoke-tests/package.json
@@ -4,11 +4,18 @@
   "main": "index.js",
   "license": "MIT",
   "scripts": {
-    "test": "yarn faltest faltest/**/*-test.js"
+    "lint:js": "yarn eslint .",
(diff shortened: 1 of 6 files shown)


packages/smoke-tests/package.json
```diff
@@ -4,11 +4,18 @@
   "main": "index.js",
   "license": "MIT",
   "scripts": {
-    "test": "yarn faltest faltest/**/*-test.js"
+    "lint:js": "yarn eslint .",
+    "test": "yarn faltest faltest/**/*-test.js --disable-timeouts --timeouts-override 900000"
```
kellyselden: --disable-timeouts is only advised as a debugging tool. is --timeouts-override not enough?

NullVoxPopuli: you right, forgot disable-timeouts was in there

PR #263, opened 2019-10-16
pinning chat implementation

halawami:
Pinned chats always show even if 'hide offline users is clicked'. When pinning a chat, it directly goes to the top of the list but still under the person.

### DIFF
--- a/packages/frontend/app/models/contact.ts
+++ b/packages/frontend/app/models/contact.ts
@@ -18,7 +18,7 @@ export enum STATUS {
 
 export default class Contact extends Identity {
   @attr() onlineStatus?: STATUS;
-  @attr('boolean', { defaultValue: false }) isPinned!: boolean;
+  @attr('boolean', { defaultValue: false }) isPinned?: boolean;
 }
 
 // DO NOT DELETE: this is how TypeScript knows how to look up your models.
--- a/packages/frontend/app/models/contact.ts
+++ b/packages/frontend/app/models/contact.ts
@@ -18,7 +18,7 @@ export enum STATUS {
 
 export default class Contact extends Identity {
   @attr() onlineStatus?: STATUS;
-  @attr('boolean', { defaultValue: false }) isPinned?: boolean;
+  @attr() isPinned?: boolean;
 }
 
 // DO NOT DELETE: this is how TypeScript knows how to look up your models.
--- a/packages/frontend/app/models/contact.ts
+++ b/packages/frontend/app/models/contact.ts
@@ -18,6 +18,7 @@ export enum STATUS {
 
 export default class Contact extends Identity {
   @attr() onlineStatus?: STATUS;
+  @attr('boolean', { defaultValue: false }) isPinned!: boolean;
 }
 
 // DO NOT DELETE: this is how TypeScript knows how to look up your models.
--- a/packages/frontend/app/pods/components/pod/application/off-canvas-container/sidebar/contacts/component.ts
+++ b/packages/frontend/app/pods/components/pod/application/off-canvas-container/sidebar/contacts/component.ts
@@ -25,13 +25,11 @@ export default class ContactsSidebar extends Component<IArgs> {
   }
 
   get contacts() {
+    const sortedContacts = this.allContacts.sort(compareFunction);
     if (!this.hideOfflineContacts) {
-      return this.allContacts;
+      return sortedContacts;
     }
-
-    return this.allContacts.filter(contact => {
-      return contact.onlineStatus !== STATUS.OFFLINE;
-    });
+    return sortedContacts.filter(shouldBeInSidebar);
   }
 
   get hideOfflineContacts() {
@@ -39,7 +37,7 @@ export default class ContactsSidebar extends Component<IArgs> {
   }
 
   get offlineContacts() {
-    return this.allContacts.filter(contact => contact.onlineStatus === STATUS.OFFLINE);
+    return this.allContacts.filter(contact => !shouldBeInSidebar(contact));
   }
 
   get numberOffline() {
@@ -58,3 +56,16 @@ export default class ContactsSidebar extends Component<IArgs> {
     this.router.transitionTo('add-friend');
   }
 }
+
+const compareFunction = function(contact1: Contact, contact2: Contact) {
+  if (contact1.isPinned === contact2.isPinned) {
+    return 0;
+  } else if (contact1.isPinned) {
+    return -1;
+  }
+  return 1;
+};
+
+const shouldBeInSidebar = function(contact: Contact) {
+  return contact.onlineStatus !== STATUS.OFFLINE || contact.isPinned;
+};
--- a/packages/frontend/app/pods/components/pod/application/off-canvas-container/sidebar/contacts/contact/component.ts
+++ b/packages/frontend/app/pods/components/pod/application/off-canvas-container/sidebar/contacts/contact/component.ts
@@ -32,6 +32,12 @@ export default class SidebarContact extends Component<IArgs> {
 
   @reads('settings.hideOfflineContacts') hideOfflineContacts!: boolean;
 
+  get isPinned() {
+    const { contact } = this.args;
+
+    return contact.isPinned;
+  }
+
   get isActive() {
     const { contact } = this.args;
 
@@ -41,20 +47,19 @@ export default class SidebarContact extends Component<IArgs> {
   get shouldBeRendered() {
     const { contact } = this.args;
 
-    if (contact.id === currentUserId) {
+    let shouldRender =
+      contact.id === currentUserId ||
+      // always show if online
+      contact.onlineStatus !== STATUS.OFFLINE ||
+      // always show if there are unread messages
+      this.hasUnread ||
+      // always show if contact is pinned
+      this.isPinned;
+
+    if (shouldRender) {
       return true;
     }
-
-    // always show if online
-    if (contact.onlineStatus !== STATUS.OFFLINE) {
-      return true;
-    }
-
-    // always show if there are unread messages
-    if (this.hasUnread) {
-      return true;
-    }
-
+    
     // do not show offline contacts if configured that way
     return !this.hideOfflineContacts;
   }
@@ -84,5 +89,17 @@ export default class SidebarContact extends Component<IArgs> {
     }
 
     this.router.transitionTo('chat.privately-with', this.args.contact.id);
+  }
+
+  @action onPin() {
+    const { contact } = this.args;
+    contact.set('isPinned', !contact.isPinned);
+    contact.save();
+  }
+
+  get canBePinned() {
+    const { contact } = this.args;
+    // can't pin your own chat
+    return contact.id !== currentUserId;
   }
 }
--- a/packages/frontend/app/pods/components/pod/application/off-canvas-container/sidebar/contacts/contact/component.ts
+++ b/packages/frontend/app/pods/components/pod/application/off-canvas-container/sidebar/contacts/contact/component.ts
@@ -59,7 +59,7 @@ export default class SidebarContact extends Component<IArgs> {
     if (shouldRender) {
       return true;
     }
-    
+
     // do not show offline contacts if configured that way
     return !this.hideOfflineContacts;
   }
--- a/packages/frontend/app/pods/components/pod/application/off-canvas-container/sidebar/contacts/contact/template.hbs
+++ b/packages/frontend/app/pods/components/pod/application/off-canvas-container/sidebar/contacts/contact/template.hbs
@@ -1,23 +1,32 @@
 {{#if this.shouldBeRendered}}
-  <a
-    data-test-contact-row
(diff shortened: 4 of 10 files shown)


packages/frontend/app/pods/components/pod/application/off-canvas-container/sidebar/contacts/contact/component.ts
```diff
@@ -55,6 +61,11 @@ export default class SidebarContact extends Component<IArgs> {
       return true;
     }
 
+    // always show if contact is pinned
```
halawami: codeClimate complains about having too many return statements in this class. Do you want me to mix all the ones with 'return true'  into one if statement?

packages/frontend/app/pods/components/pod/application/off-canvas-container/sidebar/contacts/component.ts
```diff
@@ -25,12 +25,13 @@ export default class ContactsSidebar extends Component<IArgs> {
   }
 
   get contacts() {
+    var sortedContacts: Contact[];
```
NullVoxPopuli: Some notes on let vs const vs var 
https://medium.com/podiihq/javascript-variables-should-you-use-let-var-or-const-394f7645c88f

:)

that said, I should find a lint rule that discourages `var`

NullVoxPopuli: oh actually, it's below, nevermind ;)

packages/frontend/app/pods/components/pod/application/off-canvas-container/sidebar/contacts/contact/component.ts
```diff
@@ -32,6 +32,12 @@ export default class SidebarContact extends Component<IArgs> {
 
   @reads('settings.hideOfflineContacts') hideOfflineContacts!: boolean;
 
+  get getIsPinned(){
```
NullVoxPopuli: a getter doesn't need to have `get` in the name, cause there is already a `get` keyword ;)

packages/frontend/app/pods/components/pod/application/off-canvas-container/sidebar/contacts/contact/component.ts
NullVoxPopuli: naw, clarity of intent always wins of conciseness

NullVoxPopuli: however, there may be a way to convert this and still have clarity of intent, maybe...
```ts
let shouldRender = (
  contact.id === currentUserId
  // always show if online
  || contact.onlineStatus !== STATUS.OFFLINE
  // always show if there are unread messages
  || this.hasUnread
  // always show if contact is pinned
  || this.isPinned

);

```

packages/frontend/app/pods/components/pod/application/off-canvas-container/sidebar/contacts/contact/component.ts
```diff
@@ -86,4 +97,15 @@ export default class SidebarContact extends Component<IArgs> {
 
     this.router.transitionTo('chat.privately-with', this.args.contact.id);
   }
+
+  @action onPinClick() {
```
NullVoxPopuli: the action of pinning may not always be a click, it may be a spacebar or enter key press. 

maybe the action could simply be the verb `pin`?

packages/frontend/app/pods/components/pod/application/off-canvas-container/sidebar/contacts/contact/component.ts
```diff
@@ -86,4 +97,15 @@ export default class SidebarContact extends Component<IArgs> {
...
+    const { contact } = this.args;
+    contact.set('isPinned', !contact.isPinned);
+    contact.save();
+  }
+
+  get isOtherContacts() {
```
NullVoxPopuli: is there a better property name that would better describe the intent here?

packages/frontend/app/styles/components/messages.css
```diff
@@ -32,6 +32,11 @@
   }
 }
 
+a.contact-display{
```
NullVoxPopuli: since the work here is in the sidebar, should the css for the sidebar go with the sidebar styles? 
this stylesheet is intended for the message list in the chat history view

packages/frontend/app/pods/components/pod/application/off-canvas-container/sidebar/contacts/component.ts
```diff
@@ -25,12 +25,13 @@ export default class ContactsSidebar extends Component<IArgs> {
   }
 
   get contacts() {
+    const sortedContacts = this.allContacts.sort((a, b) =>
```
NullVoxPopuli: Nested ternaries can be hard to read. I should really have a lint rule for them ;)
Anyway you can extract the sortComparator to a function?
It can live in module space above or below the class

packages/frontend/app/pods/components/pod/application/off-canvas-container/sidebar/contacts/component.ts
```diff
@@ -25,12 +25,13 @@ export default class ContactsSidebar extends Component<IArgs> {
...
+      a.isPinned === b.isPinned ? 0 : a.isPinned ? -1 : 1);
     if (!this.hideOfflineContacts) {
-      return this.allContacts;
+      return sortedContacts;
     }
-
-    return this.allContacts.filter(contact => {
-      return contact.onlineStatus !== STATUS.OFFLINE;
+    return sortedContacts.filter(contact => {
+      return contact.onlineStatus !== STATUS.OFFLINE || contact.isPinned;
```
NullVoxPopuli: I wonder if we should extract an 'shouldBeInSidebar' helper function, an well as the opposite (for below)

packages/frontend/app/pods/components/pod/application/off-canvas-container/sidebar/contacts/contact/template.hbs
```diff
@@ -4,7 +4,7 @@
     href={{url-for 'chat.privately-with' @contact.id}}
     {{did-insert (perform this.findRelevantMessages)}}
     {{on 'click' (prevent-default this.onClick)}}
-    class={{if this.isActive 'is-active'}}
+    class={{if this.isActive 'is-active contact-display' 'contact-display'}}
```
NullVoxPopuli: You could make the value of class a string and have contact-display appear before the curlies, so it only appears once

packages/frontend/app/styles/components/sidebar.css
```diff
@@ -0,0 +1,13 @@
+a.contact-display{
+    display: inline-block;
+    width: 80%;
```
NullVoxPopuli: Why always 80%

Would something with dynamic frow be better here? Maybe something using css-grid?

NullVoxPopuli:
Looking good! Biggest thing left is testing ;)

packages/frontend/app/styles/components/sidebar.css
halawami: Isn't 80% considered dynamic where it will always be 80% of the sidebar? or is it just better/cleaner to use something like a grid?

NullVoxPopuli: 80% just seems arbitrary, wondering why you chose 80%?

but yeah, with grid, things are more concrete and predictable -- and easier to adapt to responsive designs.

NullVoxPopuli:
The git history looks a little messed up there. Can you rebase?

Here is my usual process:

```
git remote add upstream git@github.com:NullVoxPopuli/emberclear.git
git checkout master
git fetch upstream
get reset upstream/master --hard
git checkout pinchats
git rebase master
# if conflicts, resolve
git add <resolved file>
git rebase --continue
git push origin pinchats --force-with-lease

```

Generally, it's best to not merge in upstream work when your base is behind the upstream. It ends up interweaving your work in between the upstream commits, and it gets harder to maintain. ;)

halawami:
The failing check is SonarCloud, Preston told me to ignore for now because it needs a bit more tweaking.

packages/frontend/app/models/contact.ts
```diff
@@ -18,6 +18,7 @@ export enum STATUS {
 
 export default class Contact extends Identity {
   @attr() onlineStatus?: STATUS;
+  @attr('boolean', { defaultValue: false }) isPinned?: boolean;
```
NullVoxPopuli: What was the reason for the attr transform? ('boolean')?

I think this is the same:
```ts
@attr() isPinned?: boolean;
```

packages/frontend/app/styles/components/sidebar.css
```diff
@@ -0,0 +1,19 @@
+div.contact-grid {
```
NullVoxPopuli: are all these classes needed? 
maybe just apply the `grid` class to the container?, which is defined as:
```
.grid {
  @apply --grid;
}
```

packages/frontend/app/styles/components/sidebar.css
```diff
@@ -0,0 +1,19 @@
...
+}   
+
+button.pin-button{
+    background-color: transparent; 
+    border: none;
+    grid-area: pin;
+    text-align: right;
+}
+
+.pin-icon{
```
NullVoxPopuli: what color is the icon otherwise?

packages/frontend/tests/acceptance/sidebar-test.ts
```diff
@@ -68,6 +68,29 @@ module('Acceptance | Sidebar', function(hooks) {
           assert.notOk(page.sidebar.contacts.offlineCount.isVisible, 'offline count is shown');
         });
 
+        module('pinned contact are to be shown', function(hooks) {
+          hooks.beforeEach(async function() {
+            const pinButton = document.getElementsByClassName('pin-button')[0];
```
NullVoxPopuli: this should be a page-object reference. :-\

packages/frontend/tests/acceptance/sidebar-test.ts
```diff
@@ -101,6 +124,46 @@ module('Acceptance | Sidebar', function(hooks) {
           assert.equal(page.sidebar.contacts.list.length, 3, 'there are 3 contacts');
         });
 
+        module('pinned contacts are to be shown', function(hooks) {
+          hooks.beforeEach(async function() {
+            const pinButton = document.getElementsByClassName('pin-button')[0];
```
NullVoxPopuli: these should be page-object references

packages/frontend/app/styles/components/sidebar.css
NullVoxPopuli: as is, the pin button is a big large
image

NullVoxPopuli: playing around with this, it looks like if the grid-areas go away, and the containing div be given:
```css
@apply --column-grid

```

packages/frontend/app/models/contact.ts
halawami: I agree, thought we needed a default value

halawami: Agree, I though we needed a default value

packages/frontend/app/styles/components/sidebar.css
halawami: White so it doesn't show

packages/frontend/tests/acceptance/sidebar-test.ts
halawami: Do you mean something like '[data-test-sidebar-contacts-header]'?

packages/frontend/app/styles/components/sidebar.css
halawami: That is a lot nicer and cleaner!

packages/frontend/app/styles/components/sidebar.css
NullVoxPopuli: found out -- it's the purple background white text :+1:

packages/frontend/tests/acceptance/sidebar-test.ts
NullVoxPopuli: naw, that's just a selector. I mean using https://discordapp.com/channels/619942056898592768/622495839595790347/641377625172082768

packages/frontend/tests/acceptance/sidebar-test.ts
```diff
@@ -197,3 +247,17 @@ module('Acceptance | Sidebar', function(hooks) {
...
+  await waitFor(selectors.offlineCount);
+}
+
+function onlyCurrentUserIsShownTest(assert: Assert) {
+  const name = getService('currentUser')!.name!;
+  const content = page.sidebar.contacts.listText;
+
+  assert.contains(content, name);
+  assert.equal(page.sidebar.contacts.list.length, 1, 'one user in the contacts list');
+}
```
halawami: Added these functions to remove the duplication in the tests

NullVoxPopuli:
Just tested it out. It works, nice!

PR #110, opened 2019-09-04
StateMachine based multi-client workflows

NullVoxPopuli:

Issue #58, opened 2018-12-18
Implement contact verification system


NullVoxPopuli:
A conversation I had with someone:
```
I was thinking of something like this:
 - I give you my share link
 - you click my share link, which causes your public key to be sent to me
 - I see a new user show up in my list, but they are untrusted
 - I have the option to verify the user, I click verify somewhere, and I am prompted with a modal that has a qr scanner
  - I scan your qr code (this would us to able to meet up)
  - boom, you're trusted
```

NullVoxPopuli:
This could lead to some features where the user could decide to mute notification from non-trusted users, or other 'sets' of users (TBD)

PR #432, opened 2020-01-14
Contact verification

halawami:


packages/frontend/app/components/pod/contacts/contact-table/index.hbs
```diff
@@ -16,6 +17,13 @@
         <td>
           {{first-8 contact.publicKeyAsHex}}...
         </td>
+        <td>
+          <a class='button'
```
NullVoxPopuli: This may be simpler?

```hbs
<LinkTo 
  class='button'
  @route='contacts.verify-friend' 
  @model={{contact.id}} 
>
  {{t contact.verification }}
</LinkTo>
```

NullVoxPopuli: docs here, if you're curious: https://guides.emberjs.com/release/routing/linking-between-routes/

packages/frontend/app/models/identity.ts
```diff
@@ -8,11 +8,16 @@ export interface PublicKey {
...
   @attr() name!: string;
   @attr() publicKey!: Uint8Array;
+  @attr() verified!: boolean;
 
   get publicKeyAsHex() {
     return toHex(this.publicKey);
   }
 
+  get verification() {
+    return this.verified ? 'models.identity.verified' : 'models.identity.unverified';
```
NullVoxPopuli: what are these strings here?

NullVoxPopuli:
Did you see this branch: https://github.com/NullVoxPopuli/emberclear/pull/110
I think it'd be helpful to drive this kind of workflow via state machines

packages/frontend/app/models/identity.ts
halawami: Strings to display the contact verification status in the contacts table. I know you mentioned possibly using a symbol [here](https://github.com/NullVoxPopuli/emberclear/issues/58) but I thought using a string will be simpler in the beginning.

packages/frontend/app/components/pod/contacts/contact-table/index.hbs
halawami: I agree it is cleaner! I tried using it but I had my route set up incorrectly

packages/frontend/app/models/identity.ts
NullVoxPopuli: as a part of _"Separation of Concerns"_, models (data), should not have display/view concerns. It makes maintenance and testing harder. :(

halawami: Makes sense, I'll just call the function from the .hbs and act accordingly

NullVoxPopuli:
I should probably merge that code so you can use the state machines as examples. :-\

NullVoxPopuli:
@halawami using https://xstate.js.org/viz
do you know if you'd be able to model the verify contact flow as a finite state machine?
Here is an example of me trying to do the same thing for "transferring to another device" https://xstate.js.org/viz/?gist=23cf0c840cf6403ae6a3f37e9095735f

halawami:
> @halawami using https://xstate.js.org/viz
> do you know if you'd be able to model the verify contact flow as a finite state machine?
> Here is an example of me trying to do the same thing for "transferring to another device" https://xstate.js.org/viz/?gist=23cf0c840cf6403ae6a3f37e9095735f

I was just reviewing that. I will try to get something tonight and run it by you tonight/tomorrow.

halawami:
@NullVoxPopuli how does [this](https://xstate.js.org/viz/?gist=f743b56bf30dbcd0b773ecbdb82a5543) look. Note: I commented out the states in the destination because they break my site when I uncomment them for some reason.

NullVoxPopuli:
@halawami looks like a great start! I'd rename the sub state machines though, as source and destination maybe aren't as meaningful for this scenario.

Also, if you open the inspector/console, it'll tell you errors.
It won't let you save on render unless the config is valid.

But yeah, feel free to change the names of everything and add or remove states or actions to describe contact verification. I'm thinking that any method or behavior we'd want in the app should be represented in the machine.

NullVoxPopuli:
That way, when we instantiate the machine, me just pass it a bunch of functions and the UI should hopefully just work :tm:

packages/frontend/translations/fr-FR.yaml
```diff
@@ -103,6 +103,9 @@ models:
   identity:
     name: Nom
     publicKey: Clé publique
+    verification: Statut de vérification
```
NullVoxPopuli: you don't need to worry about these other languages, crowdin's automatic github integration will PR for these

PR #434, opened 2020-01-15
Feature: transfer settings to another device

NullVoxPopuli:
Supersedes: https://github.com/NullVoxPopuli/emberclear/pull/110


sukima:
The use of xstate looks good and i like where this is going. I do have Some comments that came up.

The xstate definition in this pr feels disjointed to me. I had trouble following its flow. Maybe if it were plugged into the xstate visualizer?

The switch/case seems odd in this case. First a simple object lookup would work but also xstate provides for side effect events via services. Then the logic would be declarative along with the xstate definition itself.

Let’s catch up via chat when i am not looking at this on my mobile phone.

NullVoxPopuli:
Re-designing the state machine here: https://xstate.js.org/viz/?gist=23cf0c840cf6403ae6a3f37e9095735f

NullVoxPopuli:
@halawami using https://xstate.js.org/viz
do you know if you'd be able to model the verify contact flow as a finite state machine?
Here is an example of me trying to do the same thing for "transferring to another device" https://xstate.js.org/viz/?gist=23cf0c840cf6403ae6a3f37e9095735f

NullVoxPopuli:
How do I even github: https://xstate.js.org/viz/?gist=23cf0c840cf6403ae6a3f37e9095735f

NullVoxPopuli:
Superseded by: https://github.com/NullVoxPopuli/emberclear/pull/733

PR #485, opened 2020-02-04
Experiment: Extract generic stuff to in-repo addons / learn how to use in-repo addons

NullVoxPopuli:
purpose:
 - keeping app code focused on app code
 - demonstrating what components could be extracted to an external package
 - more obvious / thorough testing of the individual components themselves


NullVoxPopuli:
Closing, because in-app add-ons can't have their own tests

PR #488, opened 2020-02-05
alpha typescript

NullVoxPopuli:


packages/frontend/config/build/babel.js
```diff
@@ -8,7 +8,9 @@ module.exports = {
 
       'ember-cli-babel': {
         includePolyfill: false,
-        disablePresetEnv: true,
+        // not all JS features used are supported by browsers
```
NullVoxPopuli: tbh, I wonder why decorators didn't cause breakages

PR #572, opened 2020-03-01
use vertical collection for contacts list

NullVoxPopuli:


NullVoxPopuli:
Use a single list for both contacts and channels?

NullVoxPopuli:
Even though quality isn't all passing on this, it needs merged. I need to totally rework the sidebar to solve this perf issue, and there are a couple fixed bugs in here than need to go out

PR #620, opened 2020-03-15
use autostash addon instead of self-implementation

NullVoxPopuli:


NullVoxPopuli:
Current bug: when sending a message that was stashed, the stashed message is not cleared in the stash service

Issue #608, opened 2020-03-10
Update Channel model to fit what was learned in prototyping


CelticMajora:
Currently, channels are store as such here packages/frontend/app/models/message/channel.ts

From the prototype, we have learned how we need channels to be structured, so this model needs to be updated. On top of that, models for votes and context chains might be added for improved readability.

PR #648, opened 2020-03-25
Changed channel to account for new structure and added helper models

CelticMajora:
Addresses issue #608


packages/frontend/app/models/vote-chain.ts
```diff
@@ -0,0 +1,18 @@
+import Model, { attr, hasMany } from '@ember-data/model';
+
+import Identity from 'emberclear/models/identity';
+
+export default class VoteChain extends Model {
+    @hasMany('identity', { async: false }) remaining!: Identity[];
+    @hasMany('identity', { async: false }) yes!: Identity[];
+    @hasMany('identity', { async: false}) no!: Identity[];
+    @attr() previousVoteKey!: Identity;
+    @attr() previousVoteChain!: VoteChain;//declare as encrypted
```
CelticMajora: When ember tries to convert data to a model, is there a way to mark something as encrypted? If not, can you declare the type as just raw data so we can could have a encrypted field with just data. @NullVoxPopuli basically when we send a vote it will be encrypted and each layer within it will also be encrypted.

NullVoxPopuli: the type is less for semantics, and more about describing the data. 

for example: over here, where Uint8Arrays are encrypted, it's still a Uint8Array: https://github.com/NullVoxPopuli/emberclear/blob/master/packages/frontend/app/workers/crypto/utils/nacl.ts#L43

if the field ever needs to be anything _other_ than _always_ encrypted, you may want to look at managing the value in a state machine.

properties should be single-purpose, and not sometimes be encrypted or sometimes not :)

> When ember tries to convert data to a model, is there a way to mark something as encrypted? 

fyi, this is the method in charge of transforming data:
https://github.com/NullVoxPopuli/emberclear/blob/master/packages/frontend/app/serializers/application.js#L59

packages/frontend/app/models/vote.ts
```diff
@@ -0,0 +1,16 @@
+import Model, { attr } from '@ember-data/model';
+
+import Identity from 'emberclear/models/identity';
+import VoteChain from './vote-chain';
+
+export default class Vote extends Model {
```
CelticMajora: Add an ID field to keep track of votes

packages/frontend/app/models/vote-chain.ts
```diff
@@ -0,0 +1,18 @@
+import Model, { attr, hasMany } from '@ember-data/model';
+
+import Identity from 'emberclear/models/identity';
+
+export default class VoteChain extends Model {
```
CelticMajora: Needs target and purpose

packages/frontend/app/models/vote.ts
NullVoxPopuli: all models have a default id field of `id`

CelticMajora: Good to know.

packages/frontend/app/models/vote-chain.ts
CelticMajora: Would you recommend storing a fully signed one and one where the layers have been decrypted and its just the data we need? We definitely need the signed one so that when a user wants to vote they can take the previously signed vote, add their own vote, and sign it.

NullVoxPopuli: > storing a fully signed one

fully signed what?

> where the layers have been decrypted and its just the data we need?

data should be fully decrypted once it it deserialized from JSON in the messaging/network layer -- I don't think it makes sense to keep things encrypted in the data/model layer - even for signature verification -- just do that when receiving the message -- though for the chain of signature's that's different -- you could keep the signatures, but you wouldn't need to re-(de|en)crypt unless you need to rebuild the chain for some reason, I think.
idk, I could be wrong. maybe you have this process documented / drawn / diagrammed somewhere?

packages/frontend/app/models/vote-chain.ts
CelticMajora: Fixed in bc2a68f

packages/frontend/app/models/vote-chain.ts
CelticMajora: We have a diagram from a presentation last quarter, but it needs an explanation to go along with it. I'll try to make a more in detail one and send it your way.

CelticMajora: So what we have decided to do is at each vote, the sender will build the vote without the signature, then hash and sign the vote(including previous vote chains) and attach that signature. That way we have the raw data to work with and traverse easily while still being able to verify the entire chain when need be by unsigning and verifying. We would need to implement the sorting since without ordering the hash could end up looking different for each person.

packages/frontend/app/models/channel.ts
```diff
@@ -1,7 +1,9 @@
 import Model, { attr, hasMany } from '@ember-data/model';
 
 import Identity from 'emberclear/models/identity';
-import Invitation from 'emberclear/models/invitation';
+import ChannelContextChain from './channel-context-chain';
+import VoteChain from './vote-chain';
+// import Invitation from 'emberclear/models/invitation';
```
NullVoxPopuli, line 6: These comments should be removed unless there is a plan to bring them back?

CelticMajora: Sounds good. Will do that around 2PM.

PR #664, opened 2020-03-30
Second attempt at adding new message type to support voting

CelticMajora:


NullVoxPopuli:
Looks like GH actions broke. Change looks fine :+1:

Issue #606, opened 2020-03-10
Ship channel messages out of message handler to demoracy module


CelticMajora:
Channel messages are currently being handled in packages/frontend/app/services/messages/handler.ts

The skeleton is there but there is no channel message handling code. Add code such that channel messages are sent off to democracy module to be handled.

PR #669, opened 2020-03-31
initial work on handling channel message types

halawami:
Addresses issue #606


CelticMajora:
1. First to be merged: #679 
2. Tie for second to be merged: #669 #738

CelticMajora:
Make sure to check if an incoming starting vote is a duplicate of an existing active vote and reject it if so.

packages/frontend/app/services/channels/find-or-create.ts
```diff
@@ -0,0 +1,128 @@
...
+
+    const { id } = standardVoteChain;
+
+    let voteChain = undefined;
+
+    try {
+      voteChain = await this.store.findRecord('voteChain', id);
+      return voteChain;
+    } catch (e) {
+      voteChain = this.store.createRecord('voteChain', {
```
CelticMajora, line 97: We forgot to put signature in StandardVoteChain

CelticMajora: We'll need to go through the process of converting back from hex string that we did whenever we were messing with the vote verifier stuff. I think we already have the helper methods for that, so it should be pretty straightforward.

packages/frontend/app/services/channels/vote-handler.ts
```diff
@@ -0,0 +1,130 @@
...
+  @service('channels/find-or-create') findOrCreator!: FindOrCreateChannelService;
+
+  public async handleChannelVote(message: Message, raw: StandardMessage) {
+    let existingChannel = await this.findOrCreator.findOrCreateChannel(raw.channelInfo);
+    if (existingChannel?.members.contains(message.sender!)) {
+      let sentVote = message.metadata as StandardVote;
+      let existingVote = await this.findOrCreator.findOrCreateVote(sentVote);
+
+      if (existingVote !== undefined) {
+        let voteChain = await this.findOrCreator.findOrCreateVoteChain(sentVote.voteChain);
```
CelticMajora, line 25: Should this be a findOrCreate? In theory, incoming votes should never be found in our system and may cause issues below.

packages/frontend/app/services/channels/vote-handler.ts
```diff
@@ -0,0 +1,130 @@
...
+
+        // when it is a new vote, add this to the channel votes
+        if (!existingChannel.activeVotes.find((vote) => vote.id === existingVote!.id)) {
+          existingChannel.activeVotes.push(existingVote);
+          existingChannel.save();
+        }
+
+        // when this is not a new vote, overwrite existing vote entry
+        if (existingVote.voteChain.id !== voteChain!.id) {
+          existingVote.voteChain = voteChain!;
```
CelticMajora, line 44: This is where the above comment could cause issues.

CelticMajora:
Waiting on #834 till we test, as having the message factories will be a life-changer when it comes to testing receiving.

packages/frontend/app/services/channels/channel-verifier.ts
```diff
@@ -16,7 +17,7 @@ export default class ChannelVerifier extends Service {
     let somethingWrongWithPastOrSupportingVote =
       !(await this.isValidChain(channel.previousChain)) ||
       !(await this.voteVerifier.isValid(channel.supportingVote)) ||
-      !this.isVoteCompletedPositive(channel.supportingVote, channel.previousChain.admin);
+      !isVoteCompletedPositive(channel.supportingVote, channel.previousChain.admin);
```
CelticMajora, line 20: Check this file again now that we wrapped votechain with vote

packages/frontend/types/emberclear/index.d.ts
```diff
@@ -119,14 +119,6 @@ declare global {
     voteChain: StandardVoteChain;
   }
 
-  interface StandardChannelContextChain {
-    id: string;
-    admin: ChannelMember;
-    members: ChannelMember[];
-    supportingVote: StandardVoteChain;
-    previousChain?: StandardChannelContextChain;
```
CelticMajora: We will still need to have previousChain be optional

packages/frontend/types/emberclear/index.d.ts
```diff
@@ -123,8 +123,8 @@ declare global {
     id: string;
     admin: ChannelMember;
     members: ChannelMember[];
-    supportingVote: StandardVote;
-    previousChain: StandardChannelContextChain;
+    supportingVote: StandardVoteChain;
+    previousChain?: StandardChannelContextChain;
```
CelticMajora: Ah, it got fixed here, nvm

packages/frontend/app/services/channels/-utils/channel-factory.ts
```diff
@@ -23,11 +23,14 @@ export function buildChannelContextChain(
     id: contextChain.id,
     admin: buildChannelMember(contextChain.admin),
     members: contextChain.members.map((member) => buildChannelMember(member)),
-    supportingVote: buildVoteChain(contextChain.supportingVote),
+    supportingVote:
+      contextChain.previousChain === undefined && contextChain.supportingVote === undefined
+        ? undefined
+        : buildVoteChain(contextChain.supportingVote!),
```
NullVoxPopuli, line 29: I think typescript is correct here, actually.
if previousChain is truthy, than buildVoteChain could be passed undefined...

packages/frontend/app/services/channels/-utils/channel-saver.ts
```diff
@@ -0,0 +1,75 @@
+import Channel from 'emberclear/models/channel';
+import Vote from 'emberclear/models/vote';
+import VoteChain from 'emberclear/models/vote-chain';
+import ChannelContextChain from 'emberclear/models/channel-context-chain';
+import Identity from 'emberclear/models/identity';
+
+export async function saveChannel(channel: Channel) {
+  console.error('before saving channel admin');
```
NullVoxPopuli, line 8: remember to remove all these consoles ;)

packages/frontend/app/services/channels/-utils/channel-saver.ts
```diff
@@ -0,0 +1,75 @@
...
+  await channel.admin.save();
+  console.error('after saving channel admin');
+
+  let membersPromises: Promise<Identity>[] = [];
+  channel.members.forEach((member) => membersPromises.push(member.save()));
```
NullVoxPopuli, line 13:
```ts
await Promise.all(channel.members.map((member) => member.save()))
```
would replace lines 12 through 14 ;)
(and for each of these promise arrays)

packages/frontend/app/services/channels/-utils/channel-saver.ts
```diff
@@ -0,0 +1,75 @@
+import Channel from 'emberclear/models/channel';
+import Vote from 'emberclear/models/vote';
+import VoteChain from 'emberclear/models/vote-chain';
+import ChannelContextChain from 'emberclear/models/channel-context-chain';
+import Identity from 'emberclear/models/identity';
+
+export async function saveChannel(channel: Channel) {
```
NullVoxPopuli, line 7: given all this asynchrony, I wonder if all of this data needs to be in so many models, saved individually? or just in a single POJO attribute on the `Channel`? would that make things easier (if everything was just on the Channel?)

Perhaps (attempted to be described as types):
Also with questions, because I copied from the models directory, and I'm not 100% sure what each thing is
```ts
type Vote = {
  remaining: string[]; // user ids
  yes: string[]; // user ids;
  no: string[]; // user ids;
  target: Identity; // what's a target? 
  action: VOTE_ACTION;
  key: Identity; // public key of the voter
  previousVoteChain?: VoteChain;
  signature!: Uint8Array;
}
type VoteChain = {
  yes: Vote[];
  no: Vote[];
  remaining: Vote[];
  target: unknown; // what's a target?
  key: unknown; // what's a key? / which / who's key?
  previousChain: VoteChain;
}
// looking at just the two types above? is there any redundant information? 
// I'm not sure, but just by looking at the properties, it might kinda look like it?
// idk. I could have converted from the models wrong, too

class Channel extends Model {
  // ...existing stuff

  @attr members: string[] = []; // ids of the members
  @attr admin!: string; // id;
  @attr supportingVote?: VoteChain; // semantically, what does "Supporting Vote" mean?
  @attr previousChain?: VoteChain;
}
```

packages/frontend/app/services/channels/-utils/channel-saver.ts
CelticMajora: Will do. We are still in the process of debugging the tests, so once we remodel the models, we will continue

packages/frontend/app/services/channels/-utils/channel-saver.ts
CelticMajora: Ah, yeah, that works better.

packages/frontend/app/services/channels/-utils/channel-saver.ts
CelticMajora: I think with the new models, this may be fixed. Because of how all the relations work according to the ember docs, when you update your objects I believe it updates the other side of the relationship as well. I can step you through the logic behind the models on a call if you'd like, but I do think that we likely need all of them except for maybe Vote??? A lot of what we did in the model structuring is because it needs to be that way to be able to verify that contexts haven't been tampered with.

packages/frontend/app/services/channels/-utils/channel-factory.ts
CelticMajora: I need to figure this one out. The issue is that the only valid cases are when both previousChain and supportingVote don't exist or when both do exist. Having one exist and the other not is invalid.

packages/frontend/app/services/channels/-utils/channel-saver.ts
CelticMajora: Addressed in ae248b3

packages/frontend/app/services/channels/-utils/channel-saver.ts
CelticMajora: Addressed in ae248b3

packages/frontend/app/services/channels/-utils/channel-saver.ts
CelticMajora: The Vote and VoteChain above are redundant. We use VoteChain in the system to store actual vote data and Vote is essentially used as a wrapper such that one "vote" has a constant id registered with the channel even if the chain inside is changing.

As for what target and key are for, target is the identity of the target of the VOTE_ACTION. For example, if VOTE_ACTION was add and target was NullVoxPopuli, that vote-chain would be a vote to add NullVoxPopuli to a given channel. Same idea with remove and promote. The key of a votechain is the identity of the user who voted and signed that chain. This gives us the ability to check votechains still after they are initially sent since we otherwise wouldn't have the sender information. It is used to unsign the signature for verification as well as check the the change in yes/no/remaining from the previous vote is valid.

Issue #676, opened 2020-04-01
Implement vote verification logic


CelticMajora:
When receiving a vote or verifying an entire context chain, votes must be check to make sure no tampering has occurred.

PR #679, opened 2020-04-01
Add algorithm to verify that a vote has not been tampered with

CelticMajora:
Addresses issue #676


CelticMajora:
Additional steps needed:
1. Signing NaCl wrapper
2. Hashing NaCl wrapper

packages/frontend/app/components/app/sidebar/chats/channel-form/index.ts
```diff
@@ -6,7 +6,7 @@ import { action } from '@ember/object';
 
 import ChannelManager from 'emberclear/services/channel-manager';
 import { dropTask } from 'ember-concurrency-decorators';
-import Task from 'ember-concurrency/task';
+import { taskFor } from 'emberclear/utils/ember-concurrency';
```
NullVoxPopuli, line 9: these changes shouldn't be a part of your diff, as they were changes I had made

NullVoxPopuli: try the following: 
```
git fetch upstream
git checkout master
git checkout feature-vote-verification
git rebase master
git push origin feature-vote-verification --force-with-lease
```

CelticMajora: It looks like my most recent rebase went awry

CelticMajora: I rebased my master branch with upstream master then rebased my feature branch with local master. Would that cause these issues?

CelticMajora: That works.

NullVoxPopuli: forked masters often get out of sync.
I never trust my own local copy of master and always reset from upstream when I rebase and create new branches

packages/frontend/app/services/channels/vote-verifier.ts
```diff
@@ -0,0 +1,38 @@
...
+    let voteToVerifyExpectedHash: string = undefined; // NaCl.sign.open(voteToVerify.signature, voteToVerify.key)
+
+    let voteToVerifyActualHash: string = undefined; // Hash voteToVerifyActualString
+    if (voteToVerifyActualHash !== voteToVerifyExpectedHash) {
+      return false;
+    }
+    return this.verify(voteToVerify.previousVoteChain);
+  }
+
+  private generateSortedVoteString(vote: VoteChain): string {
```
NullVoxPopuli, line 20: maybe instead, something with a little less string concatenation would work? like:

```ts
function toSortedPublicKeys(identities) {
  return identities.map(identity => identity.publicKey).sort();
}

let remaining = toSortedPublicKeys(vote.remaining);
let yeses = toSortedPublicKeys(vote.yes);
let nos = toSortedPublicKeys(vote.no);

let payload = [
  remaining,
  yeses,
  nos
  vote.target.publicKey,
  vote.action,
  vote.key.publicKey,
  vote.previousVoteChain.signature,
];

return JSON.stringify(payload);
```

and then you'd need constants for accessing this kind of object -- these are like.. the key.
```ts
REMAINING = 0;
YESES = 1;
NOS = 2;
TARGET_PUBLIC_KEY = 3;
etc
```

and then parsing this would be:

```ts
let payload = JSON.parse(stringifiedPayload);

let yeses = payload[YESES];
```

NullVoxPopuli: You may even want to initialize the array ahead of time:

```ts
let payload = new Array(8 /* size */);

// ...
payload[YESES] = toSortedPublicKeys(vote.yes);
// ...
```

CelticMajora: I like the more JSON approach. Fortunately we wouldn't need to be accessing these structures as these will be hashed and compared to the signed value provided to guarantee that the chain hasn't been tampered with.

packages/frontend/app/services/channels/vote-verifier.ts
```diff
@@ -0,0 +1,46 @@
+import Service, { inject as service } from '@ember/service';
+import VoteChain from 'emberclear/models/vote-chain';
+import CryptoConnector from '../workers/crypto';
+import WorkersService from '../workers';
+import VoteSorter from 'emberclear/services/channels/vote-sorter';
+
+export default class VoteVerifier extends Service {
+  @service workers!: WorkersService;
+  @service voteSorter!: VoteSorter;
```
NullVoxPopuli, line 9: this injection should be `@service('channels/vote-sorter') voteSorter!: VoteSorter;`

services that are namespaced need their full path specified as an arg to `@service`

CelticMajora: fixed

packages/frontend/tests/unit/service/channels/vote-verifier-test.ts
```diff
@@ -123,4 +123,94 @@ module('Unit | Service | channels/vote-verifier', function (hooks) {
...
+      let user1SigningKeys = await crypto.generateSigningKeys();
+      let user1Keys = await crypto.generateKeys();
+      user1.privateSigningKey = user1SigningKeys.privateSigningKey;
+      user1.publicSigningKey = user1SigningKeys.publicSigningKey;
+      user1.privateKey = user1Keys.privateKey;
+      user1.publicKey = user1Keys.publicKey;
+
+      let user2SigningKeys = await crypto.generateSigningKeys();
+      let user2Keys = await crypto.generateKeys();
+      user2.privateSigningKey = user2SigningKeys.privateSigningKey;
```
NullVoxPopuli, line 149: this sort of stuff can be simplified as:
```ts
let user1 = store.createRecord('user', {
  ...(await crypto.generateKeys()),
  ...(await crypto.generateSigningKeys()),
});
```

packages/frontend/tests/unit/service/channels/vote-verifier-test.ts
```diff
@@ -123,4 +123,94 @@ module('Unit | Service | channels/vote-verifier', function (hooks) {
...
+      user4.publicKey = user4Keys.publicKey;
+
+      let userToAddSigningKeys = await crypto.generateSigningKeys();
+      let userToAddKeys = await crypto.generateKeys();
+      userToAdd.privateSigningKey = userToAddSigningKeys.privateSigningKey;
+      userToAdd.publicSigningKey = userToAddSigningKeys.publicSigningKey;
+      userToAdd.privateKey = userToAddKeys.privateKey;
+      userToAdd.publicKey = userToAddKeys.publicKey;
+
+      firstVote.yes = [user1];
```
NullVoxPopuli, line 175: same here:
```ts
let firstVote = store.createRecord('vote-chain', {
  yes: [user1],  
  no: [],
  remaining: [user2, user3, user4],
  action: VOTE_ACTION.ADD,
  target: userToAdd,
  key: user1,
  previousVoteChain: undefined,
  signature: await crypto.sign(
    await crypto.hash(voteSorter.generatedSortedVote(/* does this actually need the whole object? or can it take a subset? I need to poke around elsewhere in the PR */),
    user1.privateSigningKey,
  ),
});
```

packages/frontend/app/services/channels/vote-sorter.ts
```diff
@@ -0,0 +1,29 @@
+import Service from '@ember/service';
+import VoteChain from 'emberclear/models/vote-chain';
+import Identity from 'emberclear/models/identity';
+
+export default class VoteSorter extends Service {
```
NullVoxPopuli, line 5: Because this service doesn't manage any state, I think its contents could be migrated to helper functions that would just be used by the VoteChains

CelticMajora: The purpose of having it here was that the action of creating a vote will also need to use this in order to create new votes. That being said, it may be a thing where we can have a single vote manager service that handle creation and verification.

NullVoxPopuli: but a service manages shared state in the app.

if there is no state to manage, and it's just a function that receives input and returns some output, this could be sufficient:

```ts
export function nameOfFunction(inputs) {
  return theTheing;
}
```
and then you can 
```ts
import { nameOfFunction } from 'path/to/the-function';
```

but if there is a plan to have persistent state here later, then ignore me :D

CelticMajora: In that case, it could be moved elsewhere. Would it still go in the services package then?

CelticMajora: Since we are relying on the WorkersService, I think it needs to stay a service technically even though it isn't really holding state. The WorkersService is how we can use the crypto needed for the verification.

NullVoxPopuli: I,m not sure if that makes sense? As is, these functions can exist in module-space, and anyone can import them. No D.I. needed ;)

NullVoxPopuli: > Would it still go in the services package then?

There are a couple other examples of utility modules in the services sub folders. You could use those for examples?

CelticMajora: Given that they could just be imported as functions, how would we use the WorkersService? Would users of the function have to inject it themselves?

NullVoxPopuli: it's pretty much the same, but conveys better semantics.

instead of `@service('channels/vote-sorter') sorter;`

you'd do 
```ts
import { generateSortedVote as toSortedVoteChain } from 'wherever-the-module-lives';
```

packages/frontend/app/services/channels/vote-verifier.ts
```diff
@@ -0,0 +1,62 @@
+import Service, { inject as service } from '@ember/service';
+import VoteChain from 'emberclear/models/vote-chain';
+import CryptoConnector from '../workers/crypto';
+import WorkersService from '../workers';
+import VoteSorter from './vote-sorter';
+
+export default class VoteVerifier extends Service {
+  @service workers!: WorkersService;
+  @service('channels/vote-sorter') voteSorter!: VoteSorter;
```
NullVoxPopuli, line 9: ✂️

packages/frontend/app/services/channels/vote-verifier.ts
```diff
@@ -0,0 +1,62 @@
+import Service, { inject as service } from '@ember/service';
+import VoteChain from 'emberclear/models/vote-chain';
+import CryptoConnector from '../workers/crypto';
+import WorkersService from '../workers';
+import VoteSorter from './vote-sorter';
+
```
NullVoxPopuli, line 6: ➕ `import { toSortedVote } from './-utils';`

packages/frontend/app/services/channels/vote-sorter.ts
NullVoxPopuli: idk if this seems like toemaytoe tomahtoe, but services have more of a memory and initial runtime cost to them than modules do.

NullVoxPopuli: > how would we use the WorkersService? 

how you're using the worker service now is  fine. `VoteSorter` doesn't inject anything -- it has no dependencies.

CelticMajora: Makes sense. Fixed that.

packages/frontend/app/services/channels/-utils/vote-sorter.ts
```diff
@@ -0,0 +1,19 @@
...
+  let toReturn = [
+    toSortedPublicKeys(vote.remaining),
+    toSortedPublicKeys(vote.yes),
+    toSortedPublicKeys(vote.no),
+    vote.target.publicKey,
+    vote.action,
+    vote.key.publicSigningKey,
+    vote.previousVoteChain?.signature,
+  ];
+  return new TextEncoder().encode(JSON.stringify(toReturn));
```
NullVoxPopuli, line 14: You may be interested in this file: https://github.com/NullVoxPopuli/emberclear/blob/master/packages/frontend/app/utils/string-encoding.ts#L19

NullVoxPopuli: You may be interested in this file: https://github.com/NullVoxPopuli/emberclear/blob/master/packages/frontend/app/utils/string-encoding.ts#L19

CelticMajora: The nacl hasher and signing unfortunately needs to take in Uint8Arrays, so I think we will need to encode them into that regardless.

CelticMajora:
Right now, the github actions chrome tests are failing on the vote verifier tests. They are passing locally however, so we need to dig into why that could be.

packages/frontend/app/services/channels/-utils/vote-sorter.ts
CelticMajora: We could extract the object->Uint8Array into that other file though, so it could be universal across the board.

packages/frontend/app/services/channels/vote-verifier.ts
```diff
@@ -0,0 +1,61 @@
...
+    );
+
+    if (!this.checkVoteHashes(voteToVerifyActualHash, voteToVerifyExpectedHash)) {
+      return false;
+    }
+
+    if (voteToVerify.previousVoteChain === undefined) {
+      return true;
+    }
+
```
GreenMobius, line 29: should we check whether someone voted for themselves?

example: i vote yes for myself and Caleb. There's no previous and I sign it and put my key in. Passes algorithm. I can add a test to show.

GreenMobius: i'll PR a test in.

CelticMajora: Checking that the contents of a vote are valid will happen outside of this. When we receive a vote, we first check that the vote we are given is a legal vote. Then we will pass it off to the verifier to check for tampering of past votes to get to the current state.

CelticMajora: So that part will be processed in the vote handler, which will still have the sender's information.

CelticMajora:
1. First to be merged: #679 
2. Tie for second to be merged: #669 #738

packages/frontend/tests/unit/service/channels/utils/-vote-sorter-test.ts
```diff
@@ -0,0 +1,81 @@
...
+import { equalsUint8Array } from 'emberclear/utils/uint8array-equality';
+
+module('Unit | Service | channels/utils/vote-sorter', function (hooks) {
+  setupTest(hooks);
+  clearLocalStorage(hooks);
+
+  module('vote is sorted properly', function () {
+    test('when ran', function (assert) {
+      const store = getStore();
+      let yes1 = store.createRecord('identity', {
```
NullVoxPopuli, line 20: would it be better to do:
```ts
import { buildContact } from 'emberclear/tests/helpers/factories/contact-factory';
// this may need to be modified to add the signing keys
import { setupCurrentUser } from 'emberclear/tests/helpers';

module('Unit ...', function (hooks) {
  setupTest(hooks);
  clearLocalStorage(hooks);
  setupCurrentUser(hooks);
});
// ...

test('when ran', function (assert) {
  let yes1 = buildContact();
  let yes2 = buildContact();
  let no1 = buildContact();
  let no2 = buildContact();
  let remaining1 = buildContact();
  let remaining2 = buildContact();

  let currentUser = getService('current-user').record;

});

```

packages/frontend/tests/unit/service/channels/utils/-vote-sorter-test.ts
```diff
@@ -0,0 +1,81 @@
...
+        yes: [yes1, yes2],
+        no: [no1, no2],
+        remaining: [remaining2, remaining1],
+      });
+
+      let result = convertUint8ArrayToObject(generateSortedVote(currentVote));
+      assert.ok(
+        result[0].every(
+          (current: Uint8Array, index: number, array: Uint8Array[]) =>
+            !index || array[index - 1] <= current
```
NullVoxPopuli, line 59: what's this doing?

packages/frontend/tests/unit/service/channels/utils/-vote-sorter-test.ts
```diff
@@ -0,0 +1,81 @@
...
+      assert.ok(
+        result[2].every(
+          (current: Uint8Array, index: number, array: Uint8Array[]) =>
+            !index || array[index - 1] <= current
+        )
+      );
+
+      assert.ok(equalsUint8Array(result[3], currentUser.publicKey));
+      assert.equal(result[4], VOTE_ACTION.ADD);
+      assert.ok(equalsUint8Array(result[5], currentUser.publicSigningKey));
```
NullVoxPopuli, line 77: instead of using magic number indexes here, could you define something like this 
```ts
// I chose this file, because it creates this type of object
// packages/frontend/app/services/channels/-utils/vote-sorter.ts
export const VOTE = {
  remaining: 0,
  yes: 1,
  no: 2,
  submittersPublicKey: 3, // ? I who owns this?
  action: 4,
  submittersPublicSigningKey: 5,
  previousChainSignature: 6,
};

// ......
```
and the use those index-constants here? helps with readability / semantics and such.

example:
```ts
assert.ok(equalsUint8Array(
  result[VOTE.previousChainSignature],
  firstVote.signature
));
```

packages/frontend/tests/unit/service/channels/utils/-vote-sorter-test.ts
CelticMajora: Guaranteeing that the public keys are in numerical order in the yes, no, and remaining arrays.

packages/frontend/tests/unit/service/channels/utils/-vote-sorter-test.ts
CelticMajora: Makes sense.

packages/frontend/tests/unit/service/channels/utils/-vote-sorter-test.ts
CelticMajora: We will probably need to make a buildUser() because they will need to have private signing keys in order to sign previous vote chains.

packages/frontend/tests/unit/service/channels/utils/-vote-sorter-test.ts
CelticMajora: Implemented. The submittersPublicKey you have above is the public key of the target the action of the vote is targeting. For example, if a vote has VOTE_ACTION.ADD and target <Identity of NullVoxPopuli>, that vote would be a vote to add NullVoxPopuli to a channel.

packages/frontend/tests/unit/service/channels/utils/-vote-sorter-test.ts
NullVoxPopuli: I gotchya. Can you add another argument to 'assert.ok' that says 'ensure keys or sorted' or something like that?

packages/frontend/tests/unit/service/channels/utils/-vote-sorter-test.ts
NullVoxPopuli: Sounds good 👍

packages/frontend/tests/unit/service/channels/utils/-vote-sorter-test.ts
CelticMajora: Added in 7766288

packages/frontend/app/services/channels/vote-verifier.ts
```diff
@@ -0,0 +1,48 @@
...
+export default class VoteVerifier extends Service {
+  @service workers!: WorkersService;
+  crypto?: CryptoConnector;
+
+  async verify(voteToVerify: VoteChain): Promise<boolean> {
+    this.connectCrypto();
+
+    let voteToVerifyActual: Uint8Array = generateSortedVote(voteToVerify);
+    let voteToVerifyActualHash: Uint8Array = await this.crypto!.hash(voteToVerifyActual);
```
NullVoxPopuli, line 16: Adding the exclamation point here isn't quite correct, because 'this.crypto' might be undefined still, like if connect crypto silently fails.

You'll want to have a guard condition above here and drop the exclamation ;)

NullVoxPopuli: Alternatively

The method could be dropped and the property initialized with the connector.

packages/frontend/app/utils/string-encoding.ts
```diff
@@ -22,6 +22,18 @@ export async function convertObjectToQRCodeDataURL(object: any): Promise<string>
   return await QRCode.toDataURL(string);
 }
 
+export function convertObjectToUint8Array(object: any): Uint8Array {
```
NullVoxPopuli, line 25: Try using a generic type instead of any here and in the next method ;)

packages/frontend/tests/unit/service/channels/utils/-vote-sorter-test.ts
```diff
@@ -0,0 +1,161 @@
...
+      assert.ok(keyEquals(result[VOTE_ORDERING.targetKey], currentUser.publicKey));
+      assert.equal(result[VOTE_ORDERING.action], VOTE_ACTION.ADD);
+      assert.ok(keyEquals(result[VOTE_ORDERING.voterSigningKey], currentUser.publicSigningKey));
+      assert.ok(keyEquals(result[VOTE_ORDERING.previousChainSignature], firstVote.signature));
+      assert.equal(Object.keys(result).length, 7);
+    });
+
+    test('when ran without previous vote', function (assert) {
+      const store = getStore();
+      let yes1 = store.createRecord('identity', {
```
NullVoxPopuli, line 86: This test should probs be updated, too

packages/frontend/tests/unit/service/channels/utils/-vote-sorter-test.ts
```diff
@@ -0,0 +1,161 @@
...
+      assert.equal(result[VOTE_ORDERING.action], VOTE_ACTION.ADD);
+      assert.ok(keyEquals(result[VOTE_ORDERING.voterSigningKey], currentUser.publicSigningKey));
+      assert.equal(result[VOTE_ORDERING.previousChainSignature], undefined);
+      assert.equal(Object.keys(result).length, 7);
+    });
+  });
+});
+
+function keyEquals(arr: number[], uint8array: Uint8Array): boolean {
+  if (Object.keys(arr).length !== uint8array.length) {
```
NullVoxPopuli, line 150: Object.keys on an array? Porque?

CelticMajora: Hussein had been playing around with it and the arr doesn't have the length property for some reason when we parse it into the JSON.

packages/frontend/tests/unit/service/channels/utils/-vote-sorter-test.ts
CelticMajora: Yep. Didn't catch those

CelticMajora: Addressed in 95a7735

packages/frontend/app/utils/string-encoding.ts
CelticMajora: Addressed in 95a7735

packages/frontend/tests/unit/service/channels/utils/-vote-sorter-test.ts
NullVoxPopuli: If it for realsies doesn't have a length property, it might not be an array. Screenshot this to me if you get stuck ;)

CelticMajora: Hussein will shoot you a screenshot soon. What should be happening is a Uint8Array is assigned to a field in a JSON and that JSON is transformed into a Uint8Array. To unpack that data for test, we then convert the Uint8Array to JSON and pull that value out. It has the same data as a Uint8Array, however it is lacking key properties, such as length and other array-like properties.

packages/frontend/app/utils/string-encoding.ts
CelticMajora: Realized this caused some type check errors. If we are passing a json that at its base in an array of things, how would we go about making that generic? We can't really declare an object structure because ultimately it is an array containing a few arrays and a few objects.

CelticMajora: Fixed typing errors, however that caused tests to fail. Going to look into why that is the case.

packages/frontend/app/services/channels/vote-verifier.ts
CelticMajora: Would there be a way to set up some sort of retry thing to prevent this. Ultimately since this is a verifier, I wouldn't really want to have either boolean if the method were to fail due to not being able to connect to the worker.

NullVoxPopuli: Yeah, but let's not worry about that scope creep yet. I need to figure out a global retry system for when anything goes wrong so the user can retry manually, or we automatically retry with a some backoff timer or something.

For now, let's assume verification fails if there is crypto worker

CelticMajora: Resolved in 0e75e91

packages/frontend/tests/unit/service/channels/utils/-vote-sorter-test.ts
```diff
@@ -0,0 +1,153 @@
...
+        action: VOTE_ACTION.ADD,
+        key: currentUser,
+        target: currentUser,
+        yes: [yes1, yes2],
+        no: [no1, no2],
+        remaining: [remaining2, remaining1],
+      });
+
+      let resultHex = convertUint8ArrayToObject<SortedVoteHex>(generateSortedVote(currentVote));
+      let result: SortedVote = [
```
NullVoxPopuli, line 48: Maybe a Todo for later would be to add a method that does this to the sort util?

CelticMajora: Sort of addressed in 2c7d500. If other things outside the test need it in the future it can be moved elsewhere, however I don't see anywhere else needing it.

Issue #682, opened 2020-04-02
Add wrappers for Signing and Hashing from NaCl


CelticMajora:
In order to verify votes in #679, we need to be able to sign and hash data using NaCl. Add in the wrappers in workers/crypto/utils/nacl.ts and incorporate them into the workers system such that the vote verifier can use them.

NullVoxPopuli:
These exist now

PR #686, opened 2020-04-03
add sign and open to crypto worker

halawami:
Addresses #682


packages/frontend/app/workers/crypto/utils/nacl.ts
```diff
@@ -16,6 +16,24 @@ export async function randomBytes(length: number) {
   return nacl.randomBytes(length);
 }
 
+export async function signMessage(
```
NullVoxPopuli, line 19: these 3 methods don't need to have message in the title, since they may not always be used for the full capital M Message.

PR #690, opened 2020-04-05
Adding a signing keypair to users

CelticMajora:
Addresses issue #691 

In order to do voting verification, users must have another keypair specifically for signing and opening signed messages.

### DIFF
--- a/packages/frontend/app/services/current-user.ts
+++ b/packages/frontend/app/services/current-user.ts
@@ -103,15 +103,9 @@ export default class CurrentUserService extends Service {
     name: string,
     privateKey: Uint8Array,
     publicKey: Uint8Array,
-    privateSigningKey?: Uint8Array,
-    publicSigningKey?: Uint8Array
+    privateSigningKey: Uint8Array,
+    publicSigningKey: Uint8Array
   ) {
-    if (!privateSigningKey || !publicSigningKey) {
-      let signingKeyPair = await this.crypto!.generateSigningKeys();
-      privateSigningKey = signingKeyPair.privateSigningKey;
-      publicSigningKey = signingKeyPair.publicSigningKey;
-    }
-
     const record = this.store.createRecord('user', {
       id: currentUserId,
       name,
--- a/packages/frontend/app/services/current-user.ts
+++ b/packages/frontend/app/services/current-user.ts
@@ -80,8 +80,14 @@ export default class CurrentUserService extends Service {
     return toHex(this.publicKey);
   }
 
+  get signingKeyHex(): string {
+    if(!this.publicSigningKey) return '';
+
+    return toHex(this.publicSigningKey);
+  }
+
   get shareUrl(): string {
-    const uri = `${ENV.host}/invite?name=${this.name}&publicKey=${this.uid}`;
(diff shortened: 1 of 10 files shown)


packages/frontend/app/services/current-user.ts
```diff
@@ -68,30 +80,39 @@ export default class CurrentUserService extends Service {
...
 
+  get signingKeyHex(): string {
+    if(!this.publicSigningKey) return '';
+
+    return toHex(this.publicSigningKey);
+  }
+
   get shareUrl(): string {
-    const uri = `${ENV.host}/invite?name=${this.name}&publicKey=${this.uid}`;
+    const uri = `${ENV.host}/invite?name=${this.name}&publicKey=${this.uid}&publicSigningKey=${this.signingKeyHex}`;
```
GreenMobius, line 90: why do we need to have the signing key in the URL?

GreenMobius: the public key should be enough to identify them

NullVoxPopuli, line 90: the signing key should not be a part of the share URL. it's already long enough as is
since signing isn't required for direct communication, there should be some other means of passing it to your friends -- especially as signing keys may change. Maybe the public signing key is passed along with each message, since if a  message is signed, the public key is required for verification.

CelticMajora: Now that we think about it, users never need each others public signing keys outside of votes, which will store them already, so we can remove the stuff to save them for other users altogether.

CelticMajora: Just removed this. No longer storing public signing keys as part of contacts. Will only use them when received with votes.

packages/frontend/app/services/current-user.ts
```diff
@@ -77,21 +89,30 @@ export default class CurrentUserService extends Service {
   async create(name: string): Promise<void> {
     await this.hydrateCrypto();
     const { publicKey, privateKey } = await this.crypto!.generateKeys();
+    const { publicSigningKey, privateSigningKey } = await this.crypto!.generateSigningKeys();
 
     // remove existing record
     await this.store.unloadAll('user');
 
-    await this.setIdentity(name, privateKey, publicKey);
+    await this.setIdentity(name, privateKey, publicKey, privateSigningKey, publicSigningKey);
```
CelticMajora, line 97: Add a version of this method that doesn't require these so that when you login with just your mnemonic, it generates a new set of signing keys.

CelticMajora: Added

packages/frontend/tests/workers/crypto/nacl-test.ts
```diff
@@ -19,6 +19,13 @@ module('Workers | Crypto | nacl', function () {
     assert.ok(boxKeys.privateKey);
   });
 
+  test('generateSigningKeys | works', async function (assert) {
```
NullVoxPopuli, line 22: Can you add a test that demonstrates signing / opening as well?

Note the encrypt/decrypt test a few lines below

halawami: Those tests will be in #686

NullVoxPopuli: sounds good!

PR #728, opened 2020-04-12
re-enable faltest

NullVoxPopuli:
Depends on: https://github.com/CrowdStrike/faltest/pull/278

### DIFF
--- a/.github/workflows/frontend-tests.yml
+++ b/.github/workflows/frontend-tests.yml
@@ -134,33 +134,37 @@ jobs:
 
 ##############################################################
 
-#   tests_e2e:
-#     name: E2E Tests
-#     runs-on: ubuntu-latest
-#     timeout-minutes: 15
+  tests_e2e:
+    name: E2E Tests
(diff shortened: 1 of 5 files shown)


NullVoxPopuli:
This needs to be changed to get the deploy comment URL: https://github.com/NullVoxPopuli/emberclear/blob/master/packages/smoke-tests/tests/smoke-test.js#L21

Or maybe the URL can be uploaded as an artifact?

NullVoxPopuli:
What are 'outputs' https://github.com/nwtgck/actions-netlify/blob/develop/README.md#outputs

NullVoxPopuli:
https://github.com/nwtgck/actions-netlify/blob/develop/.github/workflows/test.yml#L30

NullVoxPopuli:
Looks like I have to dump the output to an artifact

https://stackoverflow.com/questions/59175332/using-output-from-a-previous-job-in-a-new-one-in-a-github-action

PR #667, opened 2020-03-30
Handle text commands

amgassert:
Closes #604

PR #735, opened 2020-04-13
Sidebar Voting UI

GreenMobius:
<s>Depends on #667.</s>


NullVoxPopuli:
This may be of interest (for rebasing)
```shell
git fetch upstream
git checkout master
git checkout <branch>
git rebase master
git push origin <branch> --force-with-lease
```

NullVoxPopuli:
Deploy URL with feature enabled: https://5eb8c529c2f2d2a6ac286e62--emberclear.netlify.app/settings?_features=democracy-ui

packages/frontend/app/components/app/sidebar/index.ts
```diff
@@ -38,4 +41,25 @@ export default class Sidebar extends Component {
     scrollIntoViewOfParent(scrollable, lastRow);
     this.sidebar.clearUnreadAbove();
   }
+
+  @action
+  switchToContacts() {
+    this.contactsSelected = true;
```
NullVoxPopuli, line 47: Would an enum-like object be better here? Have one true state:tm: rather than potentially every tab active?

packages/frontend/app/components/app/sidebar/actions/action/index.hbs
```diff
@@ -0,0 +1,8 @@
+<div data-test-action-row class='sidebar__actions__action'>
+
+  <span data-test-action-name>
+    <span>{{or @action.header '[error]'}}</span>
```
NullVoxPopuli, line 4: Why the span within a span?

packages/frontend/app/components/app/sidebar/actions/action/index.ts
```diff
@@ -0,0 +1,10 @@
+import Component from '@glimmer/component';
+import Action from 'emberclear/models/action';
+
+interface IArgs {
+  action: Action;
+}
+
+export default class SidebarAction extends Component<IArgs> {
+  // TODO
```
NullVoxPopuli, line 9: Todo?

packages/frontend/app/components/app/sidebar/actions/index.hbs
```diff
@@ -0,0 +1,20 @@
+<nav class='actions' style='height: 100%; overflow-y: auto;' data-test-actions-list>
+
+	<div data-test-sidebar-actions-header class='menu-label'>
```
NullVoxPopuli, line 3: Looks like indentation got goofy here

packages/frontend/app/components/app/sidebar/actions/response-panel/index.hbs
```diff
@@ -0,0 +1,11 @@
+<button class='button button-xs' type='button' {{on 'click' this.yes}}>
+	<span class='icon'>{{fa-icon 'check'}}</span>
```
NullVoxPopuli, line 2: Indentation should be two spaces here instead of tab?
(I think the .editorconfig specifics this?)

packages/frontend/app/components/app/sidebar/index.hbs
```diff
@@ -1,13 +1,80 @@
 <aside ...attributes>
   <div>
+    {{did-insert this.switchToContacts}}
```
NullVoxPopuli, line 3: I need to remove did-insert, actually. But this can be replaced by just setting the default values on the class

packages/frontend/app/components/app/sidebar/index.hbs
```diff
@@ -1,13 +1,80 @@
...
+    {{#if (has-feature-flag 'democracy-ui')}}
```
NullVoxPopuli, line 4: Wooooo featue flags!

packages/frontend/app/components/app/sidebar/index.ts
NullVoxPopuli: Maybe a 'switchToTab' action?

packages/frontend/app/models/identity.ts
```diff
@@ -24,8 +24,7 @@ export default class Identity extends Model implements Partial<PublicKey> {
 
   // Needed otherwise this regularly invalidates
   // TODO: will the public key ever change? who knows
-  // eslint-disable-next-line
-  @computed()
+  @computed('publicKeyAsHex')
```
NullVoxPopuli, line 27: Why was this changed?

packages/frontend/app/components/app/sidebar/index.ts
```diff
@@ -20,6 +20,9 @@ export default class Sidebar extends Component {
   @alias('sidebar.hasUnreadBelow') hasUnreadBelow!: boolean;
   @reads('currentUser.name') name?: string;
   @reads('currentUser.isLoggedIn') isLoggedIn!: boolean;
+  @alias('sidebar.contactsSelected') contactsSelected!: boolean;
```
NullVoxPopuli, line 23: Do these properties exist on the sidebar? I didn't see any sidebar changes :/

packages/frontend/app/components/app/sidebar/actions/response-panel/index.ts
```diff
@@ -0,0 +1,26 @@
...
+import Channel from 'emberclear/models/channel';
+
+interface IArgs {
+  channel: Channel;
+}
+
+export default class ResponsePanel extends Component<IArgs> {
+  @action
+  yes() {
+    alert('voting yes!');
```
CelticMajora, line 14: I'm assuming that these will eventually call the services to send votes. Because a response panel will in theory be tied to a single vote, would it be possible to pass that vote as a parameter so it is easier to manage? Also maybe add TODOs for where the code for sending votes will be needed.

packages/frontend/app/components/app/sidebar/index.ts
CelticMajora: I would also think this would be the way to go, as this is essentially a state switcher between three states.

packages/frontend/config/build/workers.js
```diff
@@ -5,7 +5,7 @@ const babel = require('rollup-plugin-babel');
 const resolve = require('@rollup/plugin-node-resolve');
 const { terser } = require('rollup-plugin-terser');
 const filesize = require('rollup-plugin-filesize');
-const visualizer = require('rollup-plugin-visualizer');
+// const visualizer = require('rollup-plugin-visualizer');
```
CelticMajora, line 8: If this is ultimately moving to below, I think it would be best to remove the comment. Idk the standard for this sort of thing in ember though, so that's maybe a better question for @NullVoxPopuli

packages/frontend/app/components/app/sidebar/index.ts
GreenMobius: we tried using an enum but couldn't get it working with handlebars

NullVoxPopuli: how do you mean?

packages/frontend/app/components/app/off-canvas/-page.ts
```diff
@@ -45,8 +46,16 @@ export const page = create({
     search: fillable('[data-test-sidebar-search]'),
     searchInfo: text('[data-test-search-info]'),
 
+    selectContactsTab: clickable('[data-test-tab-contacts]'),
+    selectChannelsTab: clickable('[data-test-tab-channels]'),
+    selectActionsTab: clickable('[data-test-tab-actions]'),
+
+    //TODO: turn into single identifier to test which text appears.
```
NullVoxPopuli, line 53: what does this mean?

packages/frontend/app/components/app/sidebar/actions/action/index.hbs
```diff
@@ -0,0 +1,8 @@
+<div data-test-action-row class='sidebar__actions__action'>
+
+  <span data-test-action-name>
+    {{or @action.header '[error]'}}
```
NullVoxPopuli, line 4: what is '[error]'? when would it be displayed?

packages/frontend/app/components/app/sidebar/actions/index.ts
```diff
@@ -0,0 +1,57 @@
...
+
+export default class ActionsSidebar extends Component<IArgs> {
+  @service currentUser!: CurrentUserService;
+  @service settings!: SettingsService;
+  @service router!: RouterService;
+  @service store!: StoreService;
+  @service sidebar!: SidebarService;
+
+  get allActions(): Action[] {
+    //TODO: filter to only current channel
```
NullVoxPopuli, line 32: what is this waiting on?

packages/frontend/app/components/app/sidebar/actions/index.ts
```diff
@@ -0,0 +1,57 @@
...
+      .filter((action) => action.response !== ACTION_RESPONSE.DISMISSED);
+  }
+
+  get newActions(): Action[] {
+    return this.allActions.sort(sortByOldest);
+  }
+
+  @action
+  async addAction() {
+    //TODO: Construct a new action, add to the store, and broadcast
```
NullVoxPopuli, line 45: anything holding this up?

packages/frontend/app/components/app/sidebar/actions/response-panel/index.ts
```diff
@@ -0,0 +1,29 @@
...
+
+import Channel from 'emberclear/models/channel';
+import VoteChain from 'emberclear/models/vote-chain';
+import Vote from 'emberclear/models/vote';
+
+interface IArgs {
+  vote: VoteChain;
+}
+
+//TODO: discuss user changing their mind
```
NullVoxPopuli, line 13: what's holding up implementation?

packages/frontend/app/components/app/sidebar/chats/index.hbs
```diff
@@ -1,4 +1,4 @@
-<nav class='contacts' style='height: 100%; overflow-y: hidden;' data-test-contacts-list>
```
NullVoxPopuli, line 1: why'd this change?

packages/frontend/app/components/app/sidebar/chats/list.hbs
```diff
@@ -19,6 +19,19 @@
...
       </a>
     </div>
+  {{else if (eq chat 'add-channel')}}
+    <div data-test-sidebar-channels-header class='menu-label'>
+      {{t 'ui.sidebar.channels.title'}}
+
+      <a
+        data-test-add-channel
+        class='button button-xs'
+        {{!-- TODO: link to creating channels --}} href={{url-for 'add-friend'}}
```
NullVoxPopuli, line 29: actually, is all this pending the text commands PR?

packages/frontend/app/components/app/sidebar/index.hbs
```diff
@@ -1,20 +1,126 @@
 <aside ...attributes>
+  {{#if (has-feature-flag 'democracy-ui')}}
+    <nav class='sidebar-nav'>
+      {{#if this.isTabContacts}}
```
NullVoxPopuli, line 4: all three of these blocks could be simplified to:

```hbs
<button
  data-test-tab-contacts
  id='contacts-tab'
  type='button'
  class='sidebar-tab {{if this.isTabContacts "sidebar-tab-selected"}}'
  disabled={{not this.isTabContacts}}
  {{on 'click' (fn this.switchToTab this.Tab.Contacts)}}
>
  {{t 'ui.sidebar.contacts.title'}}
</button>
```

packages/frontend/app/models/action.ts
```diff
@@ -0,0 +1,27 @@
...
+  DENIED = 'denied',
+  DISMISSED = 'dismissed',
+}
+
+export default class Action extends Model {
+  @attr() vote!: VoteChain;
+  @attr() response!: ACTION_RESPONSE;
+  @attr() timestamp!: Date;
+
+  get header() {
```
NullVoxPopuli, line 17: display responsibilities usually go in the component instead of the model (which is data focused)

packages/frontend/translations/de-DE.yaml
```diff
@@ -172,6 +172,9 @@ ui:
       title: 'Kontakte'
       none: Keine Kontakte vorhanden
       numOffline: '{num} offline'
+    actions:
```
NullVoxPopuli, line 175: anything other than English is actually going to get obliterated by automation. Crowdin, an external service, manages all the translations, and will automatically open a PR with what translations are present at: https://crowdin.com/project/emberclear

packages/frontend/app/components/app/sidebar/chats/list.hbs
GreenMobius: The text commands PR was meant as an intermediate step to allow us to get the underlying data moving/constructing right. As of now, all of these TODOs rely on some underlying method or construction that hasn't been implemented yet. We're going to have a discussion on what we want to do with that tomorrow, either before or during our meeting.

packages/frontend/app/components/app/sidebar/chats/index.hbs
GreenMobius: There was an issue that was causing the tabs to pop up inside a scroll bar with the rest of the sidebar. This led to moving the tabs outside of the div that contained the list of chats/contacts/actions. The 'height: 100%' style was causing some issues with vertical alignment, and I found that removing it and adding a grid template line in the related .css file resolved the issues.

packages/frontend/app/components/app/sidebar/actions/action/index.hbs
GreenMobius: It was added as a placeholder for if actions had an issue with displaying the action name. Debating whether to turn it into an actual error message or removing it, as actions without headers should theoretically be caught on addition.

packages/frontend/app/components/app/off-canvas/-page.ts
GreenMobius: Realized I can in fact implement this with just extra use of the feature flag.

packages/frontend/app/models/action.ts
```diff
@@ -0,0 +1,27 @@
...
+
+export enum ACTION_RESPONSE {
+  NONE = 'none',
+  APPROVED = 'approved',
+  DENIED = 'denied',
+  DISMISSED = 'dismissed',
+}
+
+export default class Action extends Model {
+  @attr() vote!: VoteChain;
```
CelticMajora, line 13: This should likely be a Vote rather than a VoteChain. Vote stays constant while its VoteChain is updated in the background. Also relationships with other models should use belongsTo rather than attr. See the other channel models to see some examples.

packages/frontend/app/models/action.ts
GreenMobius: moved to the component side of things.

client/web/emberclear/app/components/app/sidebar/actions/action/index.ts
```diff
@@ -0,0 +1,18 @@
+import Component from '@glimmer/component';
+import Action from 'emberclear/models/action';
+
+interface IArgs {
+  action: Action;
+}
+
+export default class ResponsePanel extends Component<IArgs> {
+  get header() {
```
NullVoxPopuli, line 9: this is unused -- which then makes the class empty, and then it can be deleted :D

GreenMobius: This is the wrong class name- it should be used in the same directory. I'm going to change that file and this one to make sure its used right.

NullVoxPopuli: well, this matches the above file:
image

which doesn't reference `this.header`

GreenMobius: Yeah it references the action's header, which was removed when I moved it into this class, just forgot to update that reference.

GreenMobius:
@NullVoxPopuli i made a small change at sidebar/actions/response-action/index.hbs (reference this instead of the parameter's method), sidebar/actions/response-action/index.ts (class name change), and a rename in sidebar/actions/index.hbs (rename for class name change). If you don't have any comments on that then this should be ready to merge pending @amgassert 's approval.

PR #794, opened 2020-05-02
Adding additional checks to strengthen vote verification

CelticMajora:


packages/frontend/app/services/channels/vote-verifier.ts
```diff
@@ -42,6 +44,40 @@ export default class VoteVerifier extends Service {
       workerService: this.workers,
     });
   }
+
+  private isKeyMatchingVoteDiff(vote: VoteChain): boolean {
```
NullVoxPopuli, line 48: can you add a comment block here describing what the purpose of this method is in the overall verification process?

CelticMajora: Sure thing. We were planning on marking up the two verifiers as they are a little more involved.

packages/frontend/tests/unit/service/channels/vote-verifier-test.ts
```diff
@@ -95,73 +83,361 @@ module('Unit | Service | channels/vote-verifier', function (hooks) {
...
-      let secondVote = store.createRecord('vote-chain', {
-        yes: [user1, user2],
-        no: [],
-        remaining: [user3, user4],
-        action: VOTE_ACTION.ADD,
-        target: userToAdd,
-        key: user2,
-        previousVoteChain: firstVote,
-        signature: undefined,
+      test('when first voter is still in remaining', async function (assert) {
```
NullVoxPopuli, line 114: rather than saying "in remaining" (which is in implementation detail), what does it mean to be in remaining?
An integration-style test name should describe behavior more than a particular implementation

PR #803, opened 2020-05-05
try out @use/resource w/ ember-statecharts

NullVoxPopuli:
First commit: minimal effort. Need to watch Westworld season 3 finale.
 - ~~types technically incorrect, need to reference: [this function](https://github.com/LevelbossMike/ember-statecharts/pull/241/commits/1e10811b94479a93009500fa1b790b31fb7de126#diff-0161291f3ff5b4f0f81e91781650e498R83) to see what's going on.~~
 - ~~likely doesn't work, as I don't think `useMachine` actually _only_ takes a `MachineConfig`? maybe just only a `Machine`~~
 - ~~don't know if I have access to the `context` off of `currentState`.~~
 - ~~test failures should prove this, but I have type errors right now, so they won't even run. ha~~

everything hopefully works now?

### DIFF
--- a/packages/frontend/package.json
+++ b/packages/frontend/package.json
@@ -51,8 +51,7 @@
     "tweetnacl": "^1.0.3",
     "tweetnacl-util": "^0.15.1",
     "url-parse": "^1.4.7",
-    "uuid": "^8.0.0",
-    "xstate": "^4.9.1"
+    "uuid": "^8.0.0"
   },
   "devDependencies": {
     "@babel/plugin-proposal-nullish-coalescing-operator": "^7.8.3",
@@ -159,10 +158,12 @@
     "ember-service-worker-update-notify": "^3.0.0",
     "ember-sinon": "^5.0.0",
     "ember-source": "^3.18.1",
+    "ember-statecharts": "NullVoxpopuli/ember-statecharts#use-machine",
     "ember-template-lint": "2.6.0",
     "ember-test-helpers-extra": "github:NullVoxPopuli/ember-test-helpers-extra",
     "ember-test-selectors": "^4.1.0",
     "ember-test-waiters": "^2.0.1",
+    "ember-usable": "pzuraq/ember-usable#master",
     "ember-web-app": "^4.0.0",
     "eslint": "^6.8.0",
     "eslint-config-prettier": "^6.11.0",
--- a/packages/frontend/package.json
+++ b/packages/frontend/package.json
@@ -158,7 +158,7 @@
     "ember-service-worker-update-notify": "^3.0.0",
     "ember-sinon": "^5.0.0",
     "ember-source": "^3.18.1",
-    "ember-statecharts": "NullVoxpopuli/ember-statecharts#use-machine",
+    "ember-statecharts": "LevelbossMike/ember-statecharts#use-machine",
     "ember-template-lint": "2.6.0",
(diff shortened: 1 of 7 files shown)


LevelbossMike:
`useMachine` expects a machine to be passed - we can make it smarter so that passing a config also works but right now it expects a machine.

Idea was to have a separate file that exports the machine and the component uses it configures the machine via `withConfig` and `withContext`. Advantage of this is that you can create a functioning machine in the xstate visualizer that you then can copy and paste as is to inspect it in the visualizer whenever you come back to it when you need refactoring.

packages/frontend/types/ember-statecharts.d.ts
```diff
@@ -0,0 +1,10 @@
+import { MachineConfig, Interpreter, EventType, EventObject, StateMachine, InterpreterOptions } from "xstate";
+
+export function useMachine<
+  Context = {},
+  Schema = {},
+  Events extends EventObject = TODO<any>
+>(
+  machineConfig: MachineConfig<Context, Schema, Events>,
+  options: Partial<InterpreterOptions>
+): StateMachine<Context, Schema, Events>;
```
LevelbossMike, line 10: accessing the usable that is returned from `useMachine` will give you an object of the shape `{ currentState: [State](https://xstate.js.org/api/classes/state.html), send: [Interpreter.send](https://xstate.js.org/api/classes/interpreter.html#send)  }` - not sure how to type that properly because technically it returns a getter that will give you back this state.

NullVoxPopuli:
> Advantage of this is that you can create a functioning machine in the xstate visualizer that you then can copy and paste as is to inspect it in the visualizer whenever you come back to it when you need refactoring.

I still copy paste from the viz, I just don't copy everything. :P
(Omitting the first and last lines is fairly easy)

Reason why I don't use Machine in module-space is because doing so would cause a 'side-effect', which generally makes tree shaking not possible, so I avoid it altogether.

packages/frontend/package.json
```diff
@@ -51,8 +51,7 @@
     "tweetnacl": "^1.0.3",
     "tweetnacl-util": "^0.15.1",
     "url-parse": "^1.4.7",
-    "uuid": "^8.0.0",
-    "xstate": "^4.9.1"
```
NullVoxPopuli, line 55: @LevelbossMike should ember-statecharts have xstate as a peer dependency? 
I've never known what the best sub-dep strategy is.

 - move it to the app's responsibility so that the app can be more up to date if they want
 - bundle it in the addon so that the app doesn't need to worry about

LevelbossMike: I thought about this as well. Ideally I had the idea in mind that it should be the responsibility of the host app to install it. 

With `ember-statecharts` right now though it's bundled because we make use of `ember-auto-import` to be able to use xstate directly. If you have addon code that needs to be able to do `import xstate from 'xstate'` you need to declare it as a dependency of the addon, i.e. bundle it up for usage.

If you want to make use of another xstate version you can declare it as a dependency of the host app and make use of resolutions to tell yarn (I guess that works for npm too) what version you want installed.

Problem with both approaches (responsibility of host app to install, or use resolutions to customize) is the same. You can't be sure that the xstate version you are using works with `ember-statecharts`. Right now if you use `ember-statecharts` out of the box and use the xstate version it bundles you can be sure that it works - as this is tested in `ember-statecharts`' tests.

NullVoxPopuli:
YOLO

PR #834, opened 2020-05-13
Add factory for channel message types

CelticMajora:
Service that can generate message objects for channel messages, votes, and sync requests

Must be completed before testing can be done efficiently in #669

### DIFF
--- a/packages/frontend/app/models/identity.ts
+++ b/packages/frontend/app/models/identity.ts
@@ -22,6 +22,10 @@ export default class Identity extends Model implements Partial<PublicKey> {
     return toHex(this.publicKey);
   }
 
+  get publicSigningKeyAsHex() {
+    return toHex(this.publicSigningKey);
+  }
+
   // Needed otherwise this regularly invalidates
   // TODO: will the public key ever change? who knows
   // eslint-disable-next-line
--- a/packages/frontend/app/services/channels/-utils/channel-factory.ts
+++ b/packages/frontend/app/services/channels/-utils/channel-factory.ts
@@ -32,7 +32,7 @@ export function buildChannelContextChain(
 
 export function buildChannelMember(member: Identity): ChannelMember {
   return {
-    id: member.id,
+    id: member.uid,
     name: member.name,
   };
 }
--- a/packages/frontend/app/services/channels/-utils/channel-factory.ts
+++ b/packages/frontend/app/services/channels/-utils/channel-factory.ts
@@ -34,6 +34,7 @@ export function buildChannelMember(member: Identity): ChannelMember {
   return {
     id: member.uid,
     name: member.name,
+    signingKey: member.publicSigningKeyAsHex,
   };
 }
 
--- a/packages/frontend/app/services/channels/-utils/channel-factory.ts
+++ b/packages/frontend/app/services/channels/-utils/channel-factory.ts
@@ -5,9 +5,10 @@ import Identity from 'emberclear/models/identity';
 import { toHex } from 'emberclear/utils/string-encoding';
 import ChannelContextChain from 'emberclear/models/channel-context-chain';
 
-export function buildChannelInfo(channel: Channel) {
+export function buildChannelInfo(channel: Channel): StandardMessage['channelInfo'] {
   return {
     uid: channel.id,
+    admin: buildChannelMember(channel.admin),
     name: channel.name,
     members: channel.members.map((member) => buildChannelMember(member)),
     activeVotes: channel.activeVotes.map((activeVote) => buildVote(activeVote)),
@@ -22,7 +23,7 @@ export function buildChannelContextChain(
     id: contextChain.id,
     admin: buildChannelMember(contextChain.admin),
     members: contextChain.members.map((member) => buildChannelMember(member)),
-    supportingVote: buildVote(contextChain.supportingVote),
+    supportingVote: buildVoteChain(contextChain.supportingVote),
     previousChain:
       contextChain.previousChain === undefined
         ? undefined
--- a/packages/frontend/app/services/channels/-utils/channel-factory.ts
+++ b/packages/frontend/app/services/channels/-utils/channel-factory.ts
@@ -1,0 +1,62 @@
+import Channel from 'emberclear/models/channel';
+import Vote from 'emberclear/models/vote';
+import VoteChain from 'emberclear/models/vote-chain';
+import Identity from 'emberclear/models/identity';
+import { toHex } from 'emberclear/utils/string-encoding';
+import ChannelContextChain from 'emberclear/models/channel-context-chain';
+
+export function buildChannelInfo(channel: Channel) {
+  return {
+    uid: channel.id,
+    name: channel.name,
+    members: channel.members.map((member) => buildChannelMember(member)),
+    activeVotes: channel.activeVotes.map((activeVote) => buildVote(activeVote)),
+    contextChain: buildChannelContextChain(channel.contextChain),
+  };
+}
+
+export function buildChannelContextChain(
+  contextChain: ChannelContextChain
+): StandardChannelContextChain {
+  return {
+    id: contextChain.id,
+    admin: buildChannelMember(contextChain.admin),
+    members: contextChain.members.map((member) => buildChannelMember(member)),
+    supportingVote: buildVote(contextChain.supportingVote),
+    previousChain:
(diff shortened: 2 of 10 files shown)


packages/frontend/app/services/channels/-utils/channel-factory.ts
```diff
@@ -0,0 +1,62 @@
...
+    supportingVote: buildVote(contextChain.supportingVote),
+    previousChain:
+      contextChain.previousChain === undefined
+        ? undefined
+        : buildChannelContextChain(contextChain.previousChain),
+  };
+}
+
+export function buildChannelMember(member: Identity): ChannelMember {
+  return {
```
CelticMajora, line 34: We will need to add public signing key as hex in some way

CelticMajora: Addressed by b7c2db5

packages/frontend/app/models/identity.ts
```diff
@@ -22,6 +22,10 @@ export default class Identity extends Model implements Partial<PublicKey> {
     return toHex(this.publicKey);
   }
 
+  get publicSigningKeyAsHex() {
+    return toHex(this.publicSigningKey);
```
NullVoxPopuli, line 26: what happens when someone doesn't have a publicSigningKey?

packages/frontend/app/services/channels/-utils/channel-factory.ts
```diff
@@ -0,0 +1,64 @@
...
+import Vote from 'emberclear/models/vote';
+import VoteChain from 'emberclear/models/vote-chain';
+import Identity from 'emberclear/models/identity';
+import { toHex } from 'emberclear/utils/string-encoding';
+import ChannelContextChain from 'emberclear/models/channel-context-chain';
+
+export function buildChannelInfo(channel: Channel): StandardMessage['channelInfo'] {
+  return {
+    uid: channel.id,
+    admin: buildChannelMember(channel.admin),
```
NullVoxPopuli, line 11: I don't remember, but is the admin included in the hashing / construction of the chain?

packages/frontend/types/emberclear/index.d.ts
```diff
@@ -93,6 +93,7 @@ declare global {
   interface ChannelMember {
     id: string;
     name: string;
+    signingKey: string;
```
NullVoxPopuli, line 96: public or private? (I assume public, based on context)
no need to rename, but a comment might be helpful

packages/frontend/types/emberclear/index.d.ts
```diff
@@ -101,14 +102,29 @@ declare global {
     time: string;
   }
 
-  interface ChannelInvitation {
-    invitePublicKey: string;
-    result: MemberResult[];
+  interface StandardVoteChain {
```
NullVoxPopuli, line 105: This is def out of scope, but something to keep in mind (for myself :) )

- Would it be beneficial to have the Members stored in a unified Map of some sort?
   - This would allow the remaining,yes,no arrays to just be the keys of that map.
- Do we  need the full ChannelMember data in each of remaining, yes, and no?
   - is this data duplicated somewhere?
   - is there a "source of truth"?

packages/frontend/app/services/messages/factory.ts
```diff
@@ -58,6 +60,32 @@ export default class MessageFactory extends Service {
...
+      to: to.id,
+      channelInfo: buildChannelInfo(to),
+    });
+  }
+
+  buildChannelInfoSyncFulfill(to: Identity, data: Channel) {
+    return this.build({
+      type: TYPE.INFO_CHANNEL_SYNC,
+      to: to.uid,
+      body: 'TODO make this better', //TODO decide if there is a better way to indicate request vs fulfill
```
NullVoxPopuli, line 84: you could do two different types?

packages/frontend/app/services/channels/-utils/channel-factory.ts
CelticMajora: Yes. So the Channel model has admin and members, however they are mainly for ease of access. The contextChain field of the Channel model has the most updated context, the vote supporting it, and the previous chain.

packages/frontend/types/emberclear/index.d.ts
CelticMajora: It is public. I will add in a comment.

CelticMajora: Addressed in b8fb4e8

packages/frontend/types/emberclear/index.d.ts
CelticMajora: So within a voteChain, the "source of truth" regarding who can vote on that action is the base vote. When someone creates a vote, they put all the members from their context in remaining. When somebody starts an action and sends out their vote, other members would invalidate it if the yes/no/remaining pool of members was not equivalent to their list of members in their most updated channel context.

We don't technically need the full ChannelMember data in each remaining, yes, and no. In reality you only need the public keys.

One issue that we've considered is whether or not an active vote should be invalidated if another passes or if the context of a vote should essentially stay what it was when a vote started. It gets into some murky territory, but either would work. I feel the second way is better from the user perspective because it would be obnoxious to have to keep submitting actions if other passed. That being said, there are a lot more edge cases associated with that. For example, a vote could get into an incompletable state if enough of the voting members are removed from the channel before that vote can pass or fail.

packages/frontend/app/services/messages/factory.ts
CelticMajora: That would probably make it cleaner. We'll pursue that.

CelticMajora: Addressed in 94dc994. This will require some change to #669

packages/frontend/app/models/identity.ts
CelticMajora: I believe based on the work in PR #690, every user should have a public signing key as part of their user, regardless of whether they are new or returning users.

NullVoxPopuli: Users before that PR wouldn't have signing keys ;)
(I ran in to this myself with my own login, and that's why I added the migrate method to `load`) :P

CelticMajora: With settings or mnemonic?

NullVoxPopuli: both

CelticMajora: Doesn't the new login worker generate new signing keys upon login with mnemonic? Let me look into settings login a little more.

CelticMajora: For settings, we would need to add an else statement to importFromKey here:https://github.com/NullVoxPopuli/emberclear/blob/master/packages/frontend/app/services/current-user.ts such that when a settings file doesn't have a signing key, it generates a new one upon login.

NullVoxPopuli: > Doesn't the new login worker generate new signing keys upon login with mnemonic? Let me look into settings login a little more.

you right. if you login with your mnemonic, then you get a new signing key.

so it's just the settings upload that didn't mirror that behavior. 
anywho, it's resolved now :)

NullVoxPopuli: Looks like I may have misremembered again! sorry.

This is the line I added: https://github.com/NullVoxPopuli/emberclear/blob/master/packages/frontend/app/services/current-user.ts#L137

that accounts for if you are already logged in, and your browser suddenly has the code that uses signing keys

CelticMajora: Added code for this in #839

CelticMajora: Ah, if you stay logged in, we would need to generate.

CelticMajora: I believe we would just need to add checks in the getters for public and private signing keys and generate if either do not exist. Right now if you get the record using currentUser(), like I believe all the code should, it will call that migrate and generate them if they don't exist. Only in the case where those specific getters are called will there be the possibility that they won't exist. That is if none of the code tries to access record directly.

CelticMajora: Should I add that to #839 @NullVoxPopuli ?

NullVoxPopuli: no need for right now -- keep focussed on the channel stuff ;)

PR #839, opened 2020-05-16
Logging in with settings generates signing keys if not provided

CelticMajora:


NullVoxPopuli:
looks good, but it needs a test ;)

CelticMajora:
Addressed in 196055f

packages/frontend/tests/unit/service/current-user-test.ts
```diff
@@ -1,9 +1,26 @@
...
+
+  hooks.beforeEach(() => {
+    service = getService('current-user');
+  });
+
+  test('importFromKey where privateSigningKey is not present generates signing keys', async function (assert) {
+    let keys = await generateAsymmetricKeys();
+    await service.importFromKey('name', keys.privateKey);
+    assert.notStrictEqual(await service.currentUser(), null);
+    assert.ok((await service.currentUser())?.publicSigningKey);
```
NullVoxPopuli, line 20: the `currentUser` method loads the user if the user is not present. In this test, we expect to already have a user after the import, yeah? could we use the non-async `.record` instead?

PR #853, opened 2020-05-19
Fixing the models to use relationships with one another rather than attr

CelticMajora:
We flubbed the models initially because we weren't thinking in terms of a database, but of objects.

### DIFF
--- a/packages/frontend/app/models/channel-context-chain.ts
+++ b/packages/frontend/app/models/channel-context-chain.ts
@@ -7,10 +7,10 @@ import { Channel } from 'phoenix';
 export default class ChannelContextChain extends Model {
   @belongsTo('identity', { async: false, inverse: 'adminOf' }) admin!: Identity;
   @hasMany('identity', { async: false, inverse: 'memberOf' }) members!: Identity[];
-  @belongsTo('vote-chain', { async: false }) supportingVote?: VoteChain;
-  @belongsTo('channel-context-chain', { async: false, inverse: 'parentChain' }) previousChain?: ChannelContextChain;
-  @belongsTo('channel-context-chain', { async: true, inverse: 'previousChain' }) parentChain?: ChannelContextChain;
-  @belongsTo('channel', { async: true }) composesChannel?: Channel;
+  @belongsTo('vote-chain', { async: false }) supportingVote!: VoteChain;
+  @belongsTo('channel-context-chain', { async: false, inverse: 'parentChain' }) previousChain!: ChannelContextChain;
+  @belongsTo('channel-context-chain', { async: true, inverse: 'previousChain' }) parentChain!: ChannelContextChain;
+  @belongsTo('channel', { async: true }) composesChannel!: Channel;
 }
 
 // DO NOT DELETE: this is how TypeScript knows how to look up your models.
--- a/packages/frontend/app/models/channel-context-chain.ts
+++ b/packages/frontend/app/models/channel-context-chain.ts
@@ -1,13 +1,16 @@
-import Model, { attr, hasMany } from '@ember-data/model';
+import Model, { hasMany, belongsTo } from '@ember-data/model';
 
 import Identity from 'emberclear/models/identity';
 import VoteChain from './vote-chain';
+import { Channel } from 'phoenix';
 
 export default class ChannelContextChain extends Model {
-  @attr() admin!: Identity;
-  @hasMany('identity', { async: false }) members!: Identity[];
-  @attr() supportingVote!: VoteChain;
-  @attr() previousChain!: ChannelContextChain;
+  @belongsTo('identity', { async: false, inverse: 'adminOf' }) admin!: Identity;
+  @hasMany('identity', { async: false, inverse: 'memberOf' }) members!: Identity[];
+  @belongsTo('vote-chain', { async: false }) supportingVote?: VoteChain;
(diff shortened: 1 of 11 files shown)


CelticMajora:
Currently experiencing issues in the tests where these models are created. Will dig deeper to see why they are no longer being constructed properly.

packages/frontend/app/models/channel-context-chain.ts
```diff
@@ -1,13 +1,16 @@
...
 export default class ChannelContextChain extends Model {
-  @attr() admin!: Identity;
-  @hasMany('identity', { async: false }) members!: Identity[];
-  @attr() supportingVote!: VoteChain;
-  @attr() previousChain!: ChannelContextChain;
+  @belongsTo('identity', { async: false, inverse: 'adminOf' }) admin!: Identity;
+  @hasMany('identity', { async: false, inverse: 'memberOf' }) members!: Identity[];
+  @belongsTo('vote-chain', { async: false }) supportingVote!: VoteChain;
+  @belongsTo('channel-context-chain', { async: false, inverse: 'parentChain' }) previousChain!: ChannelContextChain;
+  @belongsTo('channel-context-chain', { async: true, inverse: 'previousChain' }) parentChain!: ChannelContextChain;
```
NullVoxPopuli, line 12: why `async: true`?

packages/frontend/app/models/vote-chain.ts
```diff
@@ -9,14 +11,17 @@ export enum VOTE_ACTION {
...
+  @hasMany('identity', { async: false, inverse: 'stillRemainingIn' }) remaining!: Identity[];
+  @hasMany('identity', { async: false, inverse: 'votedYesIn' }) yes!: Identity[];
+  @hasMany('identity', { async: false, inverse: 'votedNoIn' }) no!: Identity[];
+  @belongsTo('identity', { async: false, inverse: 'targetOfVote' }) target!: Identity;
   @attr() action!: VOTE_ACTION;
-  @attr() key!: Identity;
-  @attr() previousVoteChain?: VoteChain;
+  @belongsTo('identity', { async: false, inverse: 'voterOf' }) key!: Identity;
+  @belongsTo('vote-chain', { async: false, inverse: 'parentChain' }) previousVoteChain!: VoteChain;
+  @belongsTo('vote-chain', { async: true, inverse: 'previousVoteChain' }) parentChain!: VoteChain;
```
NullVoxPopuli, line 21: why async: true?

packages/frontend/app/services/channels/-utils/channel-factory.ts
```diff
@@ -56,7 +57,7 @@ export function buildVoteChain(voteChain: VoteChain): StandardVoteChain {
     action: voteChain.action,
     key: buildChannelMember(voteChain.key),
     previousVoteChain:
-      voteChain.previousVoteChain === undefined
+      voteChain.previousVoteChain === null
```
NullVoxPopuli, line 60: for falsey comparisons, null and undefined are the same, so this'd be equiv:
```ts
!voteChain.previousVoteChain ? undefined : buildVoteChain(voteChain.previousVoteChain)
```
Also
```ts
voteChain.previousVoteChain && buildVoteChain(voteChain.previousVoteChain)
```
returns the right most thing

packages/frontend/app/models/channel-context-chain.ts
CelticMajora: We set async to be true for the records that we will never be accessing, so that it doesn't try to load them whenever we load the records.

packages/frontend/app/models/vote-chain.ts
CelticMajora: Same thing here

packages/frontend/app/services/channels/-utils/channel-factory.ts
CelticMajora: I can look into that.

packages/frontend/app/models/vote-chain.ts
CelticMajora: Essentially we put these on all the back relationships that were non-existent before because there isn't really any use for them outside of structuring the records properly.

packages/frontend/app/models/channel-context-chain.ts
NullVoxPopuli: does the relationship need to be defined then? is the data ever loaded?

CelticMajora: I'm pretty sure that if you have to have the back relationships for them to be set up right, at least according to what I was reading in the ember docs earlier today.

packages/frontend/app/services/channels/-utils/channel-factory.ts
CelticMajora: Addressed in 820b499

packages/frontend/app/models/channel-context-chain.ts
CelticMajora: We were following https://guides.emberjs.com/release/models/relationships/

packages/frontend/app/models/vote-chain.ts
CelticMajora: Addressed in d005d5f

packages/frontend/app/models/channel-context-chain.ts
CelticMajora: Addressed in d005d5f

PR #893, opened 2020-05-29
Fixed how metadata is handled for votes

CelticMajora:


client/web/emberclear/app/services/messages/factory.ts
```diff
@@ -64,7 +64,7 @@ export default class MessageFactory extends Service {
     return this.build({
       type: TYPE.CHANNEL_VOTE,
       to: to.id,
-      metadata: buildVote(vote),
+      metadata: { [METADATA_VALUE.VOTE]: buildVote(vote) },
```
NullVoxPopuli, line 67: what is the type of buildVote? was it not Record-like?

NullVoxPopuli:
The change looks fine, but I'm unclear as to why it's needed?

client/web/emberclear/app/services/messages/factory.ts
halawami: buildVote returns a StandardMessage (an index type) and not record like

PR #898, opened 2020-05-29
Update emberclear URL in CONTRIBUTING.md

amgassert:


NullVoxPopuli:
What prompted this change?

amgassert:
> What prompted this change?

Roughly a few months ago, visiting the dev build at localhost stopped working. Now local builds are served at this weird fallback IP address. I assumed it was related to some change in our build setup rather than some change on my machine, as @GreenMobius had the same issue as well IIRC.

Are you still able to visit builds at localhost?

amgassert:
Okay, looks like this is related to the TLS change. http**s**://localhost:4201 works; I've updated the file accordingly.

GreenMobius:
I've also never been able to access via http, it's always needed https.

NullVoxPopuli:
Yup, https is required :/

NullVoxPopuli:
Also, thanks for updating this! I always forget to check this stuff

amgassert:
> When the server boots it tells you the address ;)

For me, that's `https://0.0.0.0:4201/`. Not sure why that value is given over `https://0.0.0.0:4201/`.

PR #922, opened 2020-06-07
see if the alpha version of ember-cli-typescript fixes the duplicaten babel plugin issue with embroider

NullVoxPopuli:


NullVoxPopuli:
I no longer need to experiment with this, as I plan on removing ember-cli-typescript, and using Babel for TS

